PR #1, opened 2017-05-23
FAD-4900 Core styles

jonambas:
Adds core styles, functions, mixins - all the baseline stuff
Adds jest


src/styles/index.scss
```diff
@@ -1,9 +1,6 @@
-// Global styles entry point, these are NOT BEM'ed
-
+// Vendor CSS
 @import 'vendor/normalize';
+@import url("//cloud.typography.com/6240112/781206/css/fonts.css");
```
jasonrhodes: What project does this use? Would like to avoid loading all those Gotham SSm font variants that are loaded in prod...

config/jest/cssTransform.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
```
jasonrhodes: what's going on with this jest stuff?

jonambas: Idk, copied over from CRA

jasonrhodes:
ran locally, lgtm

PR #56, opened 2017-09-19
Pass input name from props correctly

tough-griff:
The `name` prop on `TextField > input` was being set to undefined for all TextFields. This passes the prop through correctly.


jonambas:
👍

PR #79, opened 2017-11-09
Better help text spacing for text fields

jasonrhodes:
I adjusted the help text in the dev console, it was at 0.1666667rem and moving it to 0.5rem made it look good, I think. Converted the numbers and I think this does the right thing.


jasonrhodes:
Actually this isn't quite enough bc the line-height needs a slight bump up, too, and we should do for the rest of the input types. I can take a look later.

PR #85, opened 2017-12-04
Added onClose callback support

rajumsys:
To do sometime after selection is done.

### DIFF
--- a/src/components/Popover/Popover.js
+++ b/src/components/Popover/Popover.js
@@ -54,7 +54,10 @@ class Popover extends Component {
 
   handleClickOutside = (e) => {
     if (this.state.open && this.wrapper && !this.wrapper.contains(e.target)) {
-      this.setState({ open: false });
+      this.setState({ open: false }, () => {
+        this.props.onClose(this.state, e);
+      });
+      
     }
   }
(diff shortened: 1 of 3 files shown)


src/components/Popover/Popover.js
```diff
@@ -55,6 +55,7 @@ class Popover extends Component {
   handleClickOutside = (e) => {
     if (this.state.open && this.wrapper && !this.wrapper.contains(e.target)) {
       this.setState({ open: false });
+      this.props.onClose(this.state, e);
```
rajumsys: slightly, outdated. pushing latest

src/components/Popover/Popover.js
```diff
@@ -54,7 +54,10 @@ class Popover extends Component {
 
   handleClickOutside = (e) => {
     if (this.state.open && this.wrapper && !this.wrapper.contains(e.target)) {
-      this.setState({ open: false });
+      this.setState({ open: false }, () => {
+        this.props.onClose(this.state, e);
```
rajumsys: do we've a convention about what to pass to callback? i'm passing this (though not using it for my current use case)

PR #86, opened 2017-12-07
Fix tabs to not show random scrollbar

bkemper:
screen shot 2017-12-07 at 10 26 28 am

### DIFF
--- a/src/components/Tabs/Tabs.module.scss
+++ b/src/components/Tabs/Tabs.module.scss
@@ -39,7 +39,7 @@
     position: absolute;
     left: 0;
     right: 0;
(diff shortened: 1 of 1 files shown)


bkemper:
The only issue with the adjust is the underline is one pixel taller.

screen shot 2017-12-07 at 10 29 40 am

jonambas:
This property just needs to be removed I think:
```
overflow-x: scroll;
```

On windows machines, or with scrollbars always shown on mac, the x axis scrollbar still appears

bkemper:
@jonambas Good call.  Fixed.

PR #108, opened 2018-02-12
FAD-6124 Fixed pagination bug preventing current page change on perPa…

gwalls:
…ge change

### DIFF
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -10,6 +10,12 @@ class Pagination extends Component {
   static displayName = 'Pagination';
 
   static propTypes = {
+
+    /**
+     * Sets the current page number
+     */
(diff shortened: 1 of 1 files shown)


src/components/Pagination/Pagination.js
```diff
@@ -10,6 +10,12 @@ class Pagination extends Component {
   static displayName = 'Pagination';
 
   static propTypes = {
+
+    /**
+     * Sets the current page number
+     */
+    currentPage: PropTypes.number,
```
gwalls: I removed `initialIndex` in favor of `currentPage`. Before, we were essentially using the `initialIndex` prop to hold the current page index anyway. I also removed `isRequired`, as we were setting a default anyway. 
Note: this holds the page number and not the index; I thought it made sense to just translate that to an index internally

PR #114, opened 2018-03-06
Label for Checkbox.Group

bkemper:


src/components/Checkbox/Checkbox.module.scss
```diff
@@ -109,6 +109,14 @@
   }
 }
 
+.GroupLabel {
+  display: inline-block;
+  line-height: 18px;
+  margin-top: 3px;
```
jonambas: use `rem` values here ie `rem(3)` for 3px

src/components/Checkbox/Group.js
```diff
@@ -1,18 +1,26 @@
...
+  required: false
+};
+
 Group.propTypes = {
   children: PropTypes.oneOfType([
     PropTypes.arrayOf(PropTypes.node),
     PropTypes.node
-  ]).isRequired
+  ]).isRequired,
+  label: PropTypes.string,
```
jonambas: May want label to also work with react components, for example a string with an UnstyledLink

Issue #113, opened 2018-03-05
<Select> option values must be a string


jasonrhodes:
https://github.com/SparkPost/matchbox/blob/master/src/components/Select/Select.js#L36

When passing objects with value/label for options to the `<Select>` component, the values must be a string. They should at least be allowed to be a number as well.

```js
<Select options={[{ label: 'Admin', value: 1 }, { label: 'Basic', value: 100 }]} />
```

```shell
Warning: Failed prop type: Invalid prop `options[0]` supplied to `Select`.
```

PR #115, opened 2018-03-07
FAD-6286: resizable Snackbar, numeric Select values

ewandennis:
- Adds optional `maxWidth` prop to Snackbar for large chunks of text
 - Addresses #113


src/components/Snackbar/Snackbar.js
```diff
@@ -14,6 +14,11 @@ class Snackbar extends Component {
      */
     status: PropTypes.oneOf(['default', 'success', 'danger']),
 
+    /**
+     * Snackbar max-width in rem.
```
jonambas: Technically with a number passed through, this becomes a pixel value, which is fine

PR #119, opened 2018-04-09
FE-143 Debranding / Open Source

jonambas:
UI PR: https://github.com/SparkPost/2web2ui/pull/437
Jira Epic: https://jira.int.messagesystems.com/browse/FE-287

---

FE-143 -
- [x] Remove typography dep
- [x] Move EmptyState SVGs to UI
- [x] Repackage under Lerna
- [x] Export all MD icons under `matchbox-icons` package
- [x] Create additional color palettes
  - [x] Panel `accent` color  should be configurable
  - [x] Panel and Panel.Section `actions` plain button color should be configurable
  - [x] Button props `plain` and `primary` color should be configurable
  - [x] Tab colors
  - [x] Tag colors
  - [x] Progress bar colors

Others Issues -
- [x] Match UI autoprefixer settings
- [x] Button and UnstyledLink prop `Component` should be lowercased to `component`

Things to know:
- `Page` and `EmptyState` primaryAction defaults to `color: 'orange'` if color is not supplied. 
- Panel `action` and `actions` require `color: 'orange'` applied in 2web2ui. (Sorry, no way around this)
- `Button` `plain` prop is now `flat`. `plain` still works, but is left out of prop types
- `Button` `primary` prop now maps to `color='orange'`

Future todos:
- The story for EmptyState lacks an image now - should add one in as an example
- Button.module.scss is 2 big and I wish I just used flat colored buttons, or styled components
- Build scripts can probably be more simple
- Should eventually take advantage of Lerna dependency hoisting


packages/matchbox-icons/icons/index.js
```diff
@@ -0,0 +1,960 @@
+/* eslint-disable max-lines */
```
bkemper: So this and other utilities don't get buried, what do you think about making about this structure?

```
src/index.js
src/IconBase/
src/icons/*.js
```

packages/matchbox-icons/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "react-dom": "^16.3.2"
+  },
+  "devDependencies": {
+    "babel-cli": "^6.26.0",
+    "babel-core": "^6.26.0",
+    "babel-plugin-external-helpers": "^6.22.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-preset-node6": "^11.0.0",
+    "babel-preset-react": "^6.24.1",
```
bkemper: Need `babel-preset-es2015` for `npm run build`.

packages/matchbox-icons/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "url": "https://github.com/SparkPost/matchbox/issues"
+  },
+  "homepage": "https://github.com/SparkPost/matchbox/tree/master/packages/matchbox-icons#readme",
+  "sideEffects": false,
+  "scripts": {
+    "build": "babel-node scripts/build.js --presets=node6"
+  },
+  "dependencies": {
+    "react": "^16.3.2",
+    "react-dom": "^16.3.2"
```
bkemper: Should relax these dependencies to the actual versions needed to use the icons.

packages/matchbox-icons/icons/ArrowForward.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { createSvgIcon } from './IconBase';
+
+export default createSvgIcon(
```
bkemper: Is it worth making this functional and using a base icon component that accepts the path as a prop?

jonambas: Hmm not sure, I followed material-ui's implementation: https://github.com/mui-org/material-ui/tree/v1-beta/packages/material-ui-icons

Only difference is `IconBase` is in this package instead of `matchbox`. I don't see any reason why they couldn't be functional components, just not sure how to easily edit 900+ files to make this change.

packages/matchbox-icons/package.json
jonambas: other component libs seem to use `^15.3.0 || ^16.0.0`, will go with that

packages/matchbox/src/components/UnstyledLink/UnstyledLink.js
```diff
@@ -0,0 +1,63 @@
...
+  static displayName = 'UnstyledLink';
+
+  static propTypes = {
+    to: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
+    external: PropTypes.bool,
+    component: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.element
+    ]),
+    Component: PropTypes.oneOfType([
```
bkemper: Should removed like other deprecated props.

PR #120, opened 2018-04-13
FAD-6478 Modal Improvements

jonambas:


src/components/Modal/Modal.js
```diff
@@ -31,12 +59,15 @@ class Modal extends Component {
     );
 
     return (
-      <div className={modalClasses} {...rest}>
+      <div className={modalClasses} {...rest} ref={(node) => this.container = node}>
+        <div className={styles.Background} />
         <Grid center='xs' middle='xs' className={styles.Grid}>
+          <WindowEvent event='keydown' handler={this.handleKeyDown} />
+          <WindowEvent event='click' handler={this.handleOutsideClick} />
```
bkemper: These could be moved into `Content`, so they aren't always listening.

PR #123, opened 2018-05-03
FE-95: Support `errorInLabel` for `Select`

bkemper:


bkemper:
Ahh, good catch.  `Error` doesn't accept a `className` prop.

PR #130, opened 2018-06-01
Pass tab index to click handler

bkemper:


bkemper:
It might be better to pass index first, but that would be a breaking change.

bkemper:
@jonambas I reworked implementation as we discussed and added tests.

PR #132, opened 2018-06-05
Feature: Optional close button for modals

bkemper:


packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -59,12 +65,22 @@ class Modal extends Component {
...
 
     return (
-      <div className={modalClasses} {...rest} ref={(node) => this.container = node}>
+      <div className={modalClasses} onClose={onClose} {...rest} ref={(node) => this.container = node}>
         <Grid center='xs' middle='xs' className={styles.Grid}>
           <Grid.Column xs={11} md={9} xl={7}>
             <Content contentRef={(node) => this.content = node} open={open}>
               <WindowEvent event='keydown' handler={this.handleKeyDown} />
               <WindowEvent event='click' handler={this.handleOutsideClick} />
+              {showCloseButton && (
```
jonambas: If you put this outside <Content> I don't think you'll need the position absolute, maybe just a float right

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -59,12 +65,22 @@ class Modal extends Component {
...
+                <Button
+                  className={styles.CloseButton}
+                  flat
+                  onClick={onClose}
+                  outline={false}
```
jonambas: Dont think you need to specify this

packages/matchbox/src/components/Modal/Modal.module.scss
```diff
@@ -52,3 +52,10 @@
 .exited {
   pointer-events: none;
 }
+
+.CloseButton {
+  color: color(grey, 10);
```
jonambas: Not sure you want to use <Button> here, you'll need to override hover styles too right?
I'm okay with just an UnstyledLink

packages/matchbox/src/components/Modal/Modal.js
jonambas: Also you wouldn't need to pass through onClose, since it'll be considered "outside"

packages/matchbox/src/components/Modal/tests/Modal.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
jonambas: Thanks!

packages/matchbox/src/components/Modal/Modal.js
bkemper: As discussed, I tried several different ways and this seemed to make the most sense.

packages/matchbox/src/components/Modal/Modal.js
bkemper: I removed it.

packages/matchbox/src/components/Modal/Modal.module.scss
bkemper: The only hover style that was a problem was the `position: absolute` which I forces with `!important` :/.

Issue #134, opened 2018-06-13
General feedbacks


rajumsys:
_i'll write feedbacks I get while I write tests in this issue_

- Banner component accepts both `action` and `actions` props. While no issue around that, I guess it just simpler to just use only one, that is `actions`. To simulate `action`, we'll just pass it in an array `[action]`. 
- Storybook for `DatePicker`
- `id` prop for `Label` is confusing. I thought it's ID of the label. I think just `for` or currently (internally) used name `htmlFor` will make more sense.

jonambas:
+1 on the Banner props

We are getting rid of Datepicker from this library, we removed the story for now, but will need to transfer it to the UI

+1 on adjusting id for Label, will need to update the other components that use it

PR #136, opened 2018-06-14
Tests coverage for Matchbox

rajumsys:
Notes:
- I've went from top-bottom in `packages/src/components` until Tags except for components that already have tests and `Portal` (need to check how it works). 
- Created an #134 with some feedbacks


packages/matchbox/src/components/Button/tests/Button.test.js
```diff
@@ -0,0 +1,46 @@
...
+    wrapper = shallow(<Button>Hola!</Button>);
+  });
+
+  const testCases = [
+    { name: 'disabled', props: { disabled: true }},
+    { name: 'destructive', props: { destructive: true }},
+    { name: 'flat', props: { flat: true }},
+    { name: 'color navy', props: { color: 'navy' }},
+    { name: 'outline enabled', props: { outline: true }},
+    { name: 'outline orange', props: { outline: true, color: 'orange' }},
```
jonambas: outline with color is not actually supported

packages/matchbox/src/components/Datepicker/tests/Datepicker.test.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
jonambas: Can remove this file

packages/matchbox/src/components/Grid/tests/Column.test.js
```diff
@@ -0,0 +1,19 @@
...
+  beforeEach(() => {
+    wrapper = shallow(<Column><p>Grid</p></Column>);
+  });
+
+  it('renders with default props', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders with sizing prop', () => {
+    wrapper.setProps({ xs: 12, md: 6, lg: 4 });
```
jonambas: There are many other props we'd have to test here, but this is okay for now

packages/matchbox/src/components/Grid/tests/Grid.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Grid from '../Grid';
+import { shallow } from 'enzyme';
+
+describe('Grid', () => {
+  let wrapper;
+  beforeEach(() => {
+    wrapper = shallow(<Grid><p>Grid</p></Grid>);
```
jonambas: Same here, will need to test all the props eventually

packages/matchbox/src/components/Pagination/tests/Pagination.test.js
```diff
@@ -0,0 +1,36 @@
...
+
+import Pagination from '../Pagination';
+
+describe('Pagination', () => {
+  let props;
+  let wrapper;
+
+  beforeEach(() => {
+    props = {
+      currentPage: 1,
```
jonambas: Will need to add a few more tests to cover all the conditionals in this component

packages/matchbox/src/components/Table/tests/Table.test.js
```diff
@@ -0,0 +1,23 @@
...
+
+import Table from '../Table';
+
+describe('Table', () => {
+  let wrapper;
+
+  beforeEach(() => {
+    const props = {
+      title: 'Sample Table',
+      accent: true,
```
jonambas: Accent and title are not props

Will also need to test `children` pass through

PR #138, opened 2018-07-09
FE-399 - Remove datepicker from Matchbox

jamaral2:


jonambas:
in the `jest.config.json` file, can you remove the line that includes Datepicker.js 
```
  "coveragePathIgnorePatterns": [
    "index.js",
    "Datepicker.js"
  ],
```

PR #149, opened 2018-11-06
Removed requirement of "to" prop to render "Component" for `UnstyledLink`

bkemper:


jonambas:
Did you want to try this instead?
```js
render() {
    const {
      children,
      to,
      Component,
      component,
      external,
      ...rest
    } = this.props;

    const WrapperComponent = component || Component || 'a';
    const url = to ? { href: to } : null;

    return (
      <WrapperComponent
        target={external ? '_blank' : ''}
        rel={external ? 'noopener noreferrer' : ''}
        {...url}
        {...rest} >
          {children}
      </WrapperComponent>
    );
  }
```

PR #151, opened 2018-11-27
New Pager component

bkemper:
Needed for EO-423.

### DIFF
--- a/packages/matchbox/src/components/Pager/Next.js
+++ b/packages/matchbox/src/components/Pager/Next.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ArrowForward } from '@sparkpost/matchbox-icons';
+import { Button } from '../Button';
+import styles from './Pager.module.scss';
+
+const Next = ({ disabled, onClick }) => (
+  <Button className={styles.Next} disabled={disabled} onClick={onClick}>
+    <ArrowForward size={16} />
+  </Button>
+);
+
+Next.propTypes = {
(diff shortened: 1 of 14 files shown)


packages/matchbox/src/components/Pager/Next.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ArrowForward } from '@sparkpost/matchbox-icons';
+import { Button } from '../Button';
+import styles from './Pager.module.scss';
+
+const Next = ({ disabled, onClick }) => (
```
jonambas: Instead of specifying `onClick` maybe just pass through `...rest`, so we get support for `component, size, etc`

Actually can probably just pass through all props, without specifying disabled

jonambas: unless you think we don't need anything else

packages/matchbox/src/components/Pager/Pager.module.scss
```diff
@@ -0,0 +1,13 @@
+@import '../../styles/config';
+
+.Pager {
```
jonambas: This might be better suited as an inline-block element

packages/matchbox/src/components/Pager/Next.js
bkemper: I don't think `component` will be needed, but definitely size, color, etc.

bkemper:
@jonambas All review feedback has been addressed.

PR #154, opened 2018-12-12
EO-452 Prep pagination and buttons for overrides

jonambas:


stories/navigation/Pagination.stories.js
```diff
@@ -21,10 +21,12 @@ export default storiesOf('Navigation|Pagination', module)
   )))
 
   .add('with lots of pages', withInfo()(() => (
+    <Panel>
```
jonambas: Remove from story

PR #155, opened 2019-01-02
SA-384 In TextField components, the connected components will show errors when the textfield has an error

kchu1337:
The CSS for the TextField component encompasses its left and right connected components. So if the input field has errors, the left and right component field also has the error style.

### DIFF
--- a/packages/matchbox/src/components/TextField/TextField.module.scss
+++ b/packages/matchbox/src/components/TextField/TextField.module.scss
@@ -11,11 +11,6 @@
   }
 }
 
-.Error {
-  border: 1px solid color(red);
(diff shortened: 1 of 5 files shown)


packages/matchbox/src/components/TextField/TextField.module.scss
```diff
@@ -11,7 +11,7 @@
   }
 }
 
-.error .Input {
+.Error {
   border: 1px solid color(red);
```
jonambas: Doesn't look like this border is being applied

jonambas: Moving the `.Input {` rule to after `.Input {` should fix it

PR #174, opened 2019-06-06
UX-102 ComboBox Components

jonambas:
**What's changed**:
- `ComboBoxTextField`
  - New component
  - Handles styling and should be wrapped with downshift
  - Very similar to `TextField` but does not allow prefix/suffix and connectLeft/connectRight
- `ComboBoxMenu`
  - New component
  - Handles typeahead menu styling, can be used with with any downshift typeahead menu
  - (should I rename this to `TypeaheadMenu`?)
- `ComboBox`
  - ComboBox component wrapper

**How to test**
- Run through storybook with `npm run start:storybook`
- Or run the build and link locally to 2web2ui
- Test MultiSelectInput with other props from TextField

**Todo**
- [ ] Finish tests
- [ ] Rename menu component?
- [ ] Test with 2web2ui


packages/matchbox/src/components/Tag/Tag.module.scss
```diff
@@ -2,6 +2,7 @@
 
 .Tag {
   display: inline-block;
+  white-space: nowrap;
```
jonambas: This should close FE-683, needed to make the same change for this ticket

PR #183, opened 2019-07-31
UX-137 Add Slider ticks

### LINKED ISSUE #201: The Slider component does not have the relevant attributes required to properly inform screen reader users of the current set value
The `<Slider/>` component is not currently constructed 100% properly from an a11y POV.

See the following article from MDN:
[Using the aria-valuemin attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuemin_attribute)

## Issues

- [X] The slider current lacks `role="slider"` - _Note:_ Looks like I just missed this on first pass as it was already in place.
- [ ] The slider currently lacks the `aria-valuenow` attribute that informs a screen reader user of the current value of the slider

The slider *is* currently controllable with a keyboard so that's good news. Fixing the missing HTML attributes is relatively straightforward in comparison.

jonambas:
What's Changed:
- Add Slider prop `ticks`
- Resolves #201

Sample Usage
```js
<Slider value={75} ticks={{
  0: '0',
  50: '50',
  25: 'Recommended',
  100: '100',
}} />
```

### DIFF
--- a/packages/matchbox/src/components/Slider/Slider.js
+++ b/packages/matchbox/src/components/Slider/Slider.js
@@ -9,11 +9,12 @@ import { roundToPlaces, clamp } from '../../helpers/math';
 import styles from './Slider.module.scss';
 
 function Slider(props) {
-  const { defaultValue, disabled, id, max, min, onBlur, onFocus, onChange, precision, value } = props;
+  const { defaultValue, disabled, id, max, min, onBlur, onFocus, onChange, precision, ticks, value } = props;
 
   const windowSize = useWindowSize(50);
   const [sliderValue, setSliderValue] = React.useState(value || defaultValue);
   const [sliderLocation, setSliderLocation] = React.useState(0);
+  const [tickLocations, setTickLocations] = React.useState({});
   const [moving, setMoving] = React.useState();
   const sliderRef = React.useRef();
 
@@ -43,6 +44,22 @@ function Slider(props) {
       onChange(sliderValue);
     }
   }, [sliderValue, windowSize]);
+
+  // Updates tick locations when ticks or window size change
+  React.useLayoutEffect(() => {
+    if (typeof ticks === 'object') {
+      const newTicks = {};
+      const rect = getRectFor(sliderRef.current);
+      Object.keys(ticks).map((number) => {
+        const absoluteProportion = (Number(number) + Math.abs(min)) / Math.abs(min - max);
+        newTicks[number] = {
+          position: lerp(0, rect.width, absoluteProportion),
+          label: ticks[number]
+        };
+      });
+      setTickLocations(newTicks);
+    }
+  }, [ticks, windowSize]);
 
   // Event handlers
   function handleMouseDown(e) {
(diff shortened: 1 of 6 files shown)


packages/matchbox/src/components/Slider/Slider.js
```diff
@@ -44,6 +45,22 @@ function Slider(props) {
     }
   }, [sliderValue, windowSize]);
 
+  // Updates tick locations when ticks or window size change
+  React.useLayoutEffect(() => {
+    if (typeof ticks === 'object') {
```
bkemper: I think it is safe to tighten the prop types for `ticks` to `PropTypes.object` and loosen this check to existence (`if (ticks)`).

packages/matchbox/src/components/Slider/Slider.js
```diff
@@ -44,6 +45,22 @@ function Slider(props) {
...
+      const newTicks = {};
+      const rect = getRectFor(sliderRef.current);
+      Object.keys(ticks).map((number) => {
```
bkemper: `map` to `forEach`.

packages/matchbox/src/components/Slider/Slider.js
jonambas: woops `PropTypes.objectOf(PropTypes.node),`, will already throw a warning

packages/matchbox/src/components/Slider/Slider.js
```diff
@@ -44,6 +45,22 @@ function Slider(props) {
...
+        const absoluteProportion = (Number(number) + Math.abs(min)) / Math.abs(min - max);
+        newTicks[number] = {
+          position: lerp(0, rect.width, absoluteProportion),
+          label: ticks[number]
+        };
+      });
+      setTickLocations(newTicks);
```
bkemper: This feels like it should like a `useMemo`.

packages/matchbox/src/components/Slider/Slider.js
```diff
@@ -35,14 +40,35 @@ function Slider(props) {
...
+
+  // Updates tick locations when ticks or window size change
+  const tickLocations = React.useMemo(() => {
+
+    if (!ticks || !rect.width) {
+      return {};
     }
-  }, [sliderValue, windowSize]);
+
+    return Object.keys(ticks).reduce((acc, number) => {
```
bkemper: Since we are going to map over tickLocations later, why not convert it here with map instead of reduce.

```
Object.keys(ticks).map((key) => {
  // ...
  return {
    position: lerp(0, rect.width, absoluteProportion),
    label: ticks[key],
    key: tick
  }
});
```

packages/matchbox/src/components/Slider/Slider.module.scss
```diff
@@ -62,3 +62,38 @@
...
   }
 }
+
+.Tick {
+  position: absolute;
+  height: 10px;
+  width: 10px;
+  border-radius: 50%;
+  transform: translate(-3px, -3px);
+
```
bkemper: Curious, why the empty lines?

jonambas: I tend to separate box model, type, animation, interaction styles with a space like this

packages/matchbox/src/components/Slider/Slider.js
jonambas: Oh i completely overlooked this

PR #206, opened 2019-08-14
Resolves #197 tag component keyboard focusable remove button

### LINKED ISSUE #197: The Tag component remove button needs to be keyboard accessible
The `<Tag/>` component remove button is not currently keyboard accessible.

## Requirements

* The remove button should be focusable by keyboard users
* The remove button should be interactive with the enter key and spacebar

## Potential Fixes

* Update the HTML to use the `<button>` element instead of the `<a>`
* __or__ use the existing `<a>` element and add `role="button"` alongside a relevant value for `href` - `href="javascript:void(0):"` should do the trick

nicklemmon:
Resolves #197

## What Changed

* Added new `handleKeydown` method to enable spacebar handling
* Added `role="button"` and an arbitrary value to the `to` prop to allow keyboard focusing
* Added supporting test for 

## How to Test

* Go to http://localhost:9001/?selectedKind=Feedback%7CTag&selectedStory=with%20remove%20action&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel
* Click immediately before the tag component instance, hit the tab key and the button should have focus
* Hit the spacebar or enter key when focusing on the remove button and the action logger should show the mock function

### DIFF
--- a/packages/matchbox/src/components/Tag/Tag.js
+++ b/packages/matchbox/src/components/Tag/Tag.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { onKey } from '../../helpers/keyEvents';
 import { Close } from '@sparkpost/matchbox-icons';
 import { UnstyledLink } from '../UnstyledLink';
 
@@ -24,10 +25,6 @@ class Tag extends Component {
     children: PropTypes.node
   };
 
-  handleKeydown(e, callback) {
-    if (e.keyCode === 32) { callback(); }
-  }
-
   render() {
     const {
       color,
@@ -48,7 +45,7 @@ class Tag extends Component {
       ? <UnstyledLink
         className={styles.Close}
         onClick={onRemove}
-        onKeyDown={(e) => this.handleKeydown(e, onRemove)}
+        onKeyDown={onKey('space', onRemove)}
         to="javascript:void(0)"
         role="button">
         <Close size={16} />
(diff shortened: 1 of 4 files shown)


packages/matchbox/src/components/Tag/Tag.js
```diff
@@ -43,7 +47,10 @@ class Tag extends Component {
     const closeMarkup = onRemove
       ? <UnstyledLink
         className={styles.Close}
-        onClick={onRemove}>
+        onClick={onRemove}
+        onKeyDown={(e) => this.handleKeydown(e, onRemove)}
```
jonambas: We have keydown helpers in `helpers/keyEvents`. Could slim some of this down.

```js
onKeyDown={onKey('space', props.onRemove)}
```

nicklemmon: Ooh cool. I’ll definitely check those out and clean it up.

nicklemmon: Alrighty, swapped it out. Nice API there!

nicklemmon:
@jonambas for some reason, the Jest test to check for the spacebar firing the method passed in to `onRemove` doesn't work with the helper functions! But it does work otherwise. If you have any ideas I'm all ears but trying to figure that out now.

nicklemmon:
OK, got it figured out - Enzyme's simulate method does not account for `shiftKey` by default and thus returns that value as `undefined` rather than `false` by default. Kind of weird!

nicklemmon:
Nice, thanks.

So it looks like the browser default focus styles are present, which is totally kosher. We could chat about adding a custom focus style. One that I've used in the past (this is global CSS and isn't scoped, so there's that):

```
*:focus {
  outline: 1px dashed currentColor;
}
```

That ends up handling about 90% of cases. Each component might require some custom tweaking in certain scenarios, but not too bad.

For another PR, though!

PR #210, opened 2019-08-15
Issue #191 screen reader only component

### LINKED ISSUE #191: New component - ScreenReaderOnly
A `ScreenReaderOnly` component would be particularly useful - this can be used to wrap any content that will be encountered _only_ by screen reader users.

## Requirements

* The component needs to be an inline element (probably a `<span></span>`) so that it can be used in any context
* The component should only be encountered by screen reader users
* This can be a stateless, functional component

## Resources

These styles can be used to make an element visible only to screen reader users:
[Citrus Utility Library - _accessibility.scss](https://github.com/nicklemmon/citrus-util/blob/master/src/styles/4-utilities/_accessibility.scss)

nicklemmon:
Resolves #191 

## What Changed

* Added new `<ScreenReaderOnly>` component
* Added supporting unit tests
* Added example to Storybook

## How to Test

* Navigate to http://localhost:9001/?selectedKind=Utility%7CScreenReaderOnly&selectedStory=basic%20example&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel and see that certain content is in the DOM but is not available to sighted users

### DIFF
--- a/packages/matchbox/src/components/ScreenReaderOnly/ScreenReaderOnly.js
+++ b/packages/matchbox/src/components/ScreenReaderOnly/ScreenReaderOnly.js
@@ -1,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './ScreenReaderOnly.module.scss';
+
+const ScreenReaderOnly = ({ children }) => (
+  <span className={styles.ScreenReaderOnly}>
+    {children}
+  </span>
+);
+
+ScreenReaderOnly.propTypes = {
+  children: PropTypes.node.isRequired
+};
+
+export default ScreenReaderOnly;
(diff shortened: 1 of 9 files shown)


packages/matchbox/src/components/ScreenReaderOnly/ScreenReaderOnly.js
```diff
@@ -0,0 +1,15 @@
...
+import styles from './ScreenReaderOnly.module.scss';
+
+const ScreenReaderOnly = ({ children }) => (
+  <span className={styles.ScreenReaderOnly}>
+    {children}
+  </span>
+);
+
+ScreenReaderOnly.propTypes = {
+  children: PropTypes.node.isRequired
```
jonambas: Any reason anything other than a string would be used here?

nicklemmon: You *might* have other elements inside of this guy - so it could be a `<label></label>` element or even a `<legend></legend>`. You might even have hidden headings inside these as well.

PR #213, opened 2019-08-16
Update IconBase to have accessible label via a prop

nicklemmon:
## What Changed

* Added new `label` prop to `IconBase` component
* Added relevant conditional rendering for HTML attributes 

## How to Test

* Go to http://localhost:9001/?selectedKind=Icons%7Cmatchbox-icons&selectedStory=single%20icon&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel and verify that:
    + **No** `role="img"` is present on the `<svg>` tag
    + **No** `aria-label` is present on the `<svg>` tag
    + The `aria-hidden` attribute is present and set to `"true"`
* Go to http://localhost:9001/?selectedKind=Icons%7Cmatchbox-icons&selectedStory=single%20icon%20with%20a11y%20label&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel and verify that:
   + `role="img"` is present on the `<svg>` tag
   + `aria-label` is present on the `<svg>` tag with the passed in value from the `label` prop
   + The `aria-hidden` attribute is present and set to `"false"`


nicklemmon:
Whenever you get a chance, @jonambas - the CI failure is legit, yet I'm having trouble reproducing the failure locally. I think I know a way to resolve it so we can get this merged in, but it may be pointing to some sort of a larger problem with the test setup?

We can touch base sometime soon. No rush or anything.

PR #235, opened 2019-08-28
Resolve issue #218 by adding "title" prop to Table component

### LINKED ISSUE #218: The Table component should have a required prop that renders the <caption> element
The `<Table/>` component should have a required prop that accepts a string and renders the HTML `<caption>` element. To preserver the current UX, the element could be for screen readers only.

## Possible Implementation

```jsx
<Table caption="My Favorite Table"></Table>
```

Would render to...
```html
<table>
  <caption class="screen-reader-only">My Favorite Table</caption>
</table>
```

nicklemmon:
Resolves #218 

## What Changed

- Added new `title` prop to `<Table/>` component to render `<caption>`
- This new prop is required, and as a result, represents a breaking change for applications that leverage it

## How to Test

- Go to: http://localhost:9001/?selectedKind=Layout%7CTable&selectedStory=table%20components&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel and verify that the `<caption></caption>` renders
- Additionally, verify that the caption does not render visually


jonambas:
This might be something we can introduce currently without the breaking change.

Was thinking, maybe we could trigger a `console.warn` if the prop is missing, and in `4.0.0` we make the prop required

jonambas:
Moved the base branch to `rc/4.0.0`

PR #241, opened 2019-09-03
Resolves issue #239 by replacing circled icon with triangle

### LINKED ISSUE #239: Some differences between Banner component states are communicated through color alone
The "error" and "warning" variants of the `<Banner/>` component use the same icon, and unfortunately, only communicate a state difference through color alone:
Screen Shot 2019-09-03 at 11 20 43 AM
Screen Shot 2019-09-03 at 11 20 48 AM

I would advise making a very simple change - swap out the `<Error/>` icon for the `<Warning/>` icon. These two icons are very similar visually, however, the latter uses a triangle to wrap the exclamation mark that is present in both components. That is enough of a visual queue to make the two distinct from an a11y perspective, particularly when thinking about users with certain visual impairments.

nicklemmon:
Resolves #239 

## What Changed?

* Updated icon with `<Warning/>` when the 'warning' variant was using the '<Error/>` icon
* Updated styles for the warning variant specifically
* Leveraged CSS `clip-path` to build a triangle outline
* Added additional decorative element to "fill" the "i" symbol inside of `<Warning/>`

## How to Test

* Go to the ['warning' variant story of the Banner component](http://localhost:9001/?selectedKind=Feedback%7CBanner&selectedStory=Warning&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) and compare with other variants
* Note that achieving a border radius on a clip-path element is fairly difficult - if so required, we can use an inline SVG to clip the background color instead of using `clip-path` - that might look more consistent visually.

### DIFF
--- a/packages/matchbox/src/components/Banner/Banner.js
+++ b/packages/matchbox/src/components/Banner/Banner.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Error as ErrorIcon, CheckCircle, InfoOutline, Close } from '@sparkpost/matchbox-icons';
+import { Error as ErrorIcon, Warning, CheckCircle, InfoOutline, Close } from '@sparkpost/matchbox-icons';
 
 import styles from './Banner.module.scss';
 import { buttonFrom } from '../Button';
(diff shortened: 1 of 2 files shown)


jonambas:
Closing this as we're reevaluating our icon library

Issue #242, opened 2019-09-03
The modal component needs relevant aria attributes to be properly interpreted by screen readers


nicklemmon:
## Requirements

* Modal dialogs should have `role="dialog"`
* Modal dialogs should *also* have `aria-modal="true"`

That's it! Pretty easy.

jonambas:
Resolved in #244

PR #244, opened 2019-09-03
Resolve Issue #242 Add relevant aria attributes, add some explicit tests to ch…

nicklemmon:
Resolves issue #242 

## What Changed?

* Adds `role="dialog"` to the `<Modal/>` component such that screen readers interpret the component as a dialog
* Adds `aria-modal="true"` to the `<Modal/>` component to effectively "hide" other content to screen reader users when the modal is open
* Add relevant tests to check for these attributes explicitly

## How to Test

* Check the DOM for relevant attributes outlined in issue #242

### DIFF
--- a/packages/matchbox/src/components/Modal/Modal.js
+++ b/packages/matchbox/src/components/Modal/Modal.js
@@ -65,7 +65,7 @@ class Modal extends Component {
     );
 
     return (
-      <div className={modalClasses} onClose={onClose} {...rest} ref={(node) => this.container = node}>
+      <div className={modalClasses} onClose={onClose} {...rest} ref={(node) => this.container = node} role="dialog" aria-modal="true">
         <Grid center='xs' middle='xs' className={styles.Grid}>
           <Grid.Column xs={11} md={9} xl={7}>
             <Content contentRef={(node) => this.content = node} open={open}>
--- a/packages/matchbox/src/components/Modal/tests/Modal.test.js
+++ b/packages/matchbox/src/components/Modal/tests/Modal.test.js
@@ -28,6 +28,13 @@ describe('Modal', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('renders with relevant ARIA attributes', () => {
+    const wrapperProps = wrapper.props();
+
+    expect(wrapperProps.role).toEqual('dialog');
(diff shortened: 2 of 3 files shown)


packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -65,7 +65,7 @@ class Modal extends Component {
     );
 
     return (
-      <div className={modalClasses} onClose={onClose} {...rest} ref={(node) => this.container = node}>
+      <div className={modalClasses} onClose={onClose} {...rest} ref={(node) => this.container = node} role="dialog" aria-modal="true">
         <Grid center='xs' middle='xs' className={styles.Grid}>
           <Grid.Column xs={11} md={9} xl={7}>
             <Content contentRef={(node) => this.content = node} open={open}>
```
jonambas: Was wondering if we'd need `aria-hidden=true` or something similar when the modal is not open, since the modal containers are always mounted through a portal. But it looks like RTG handles mounting/unmounting children correctly for us so there is never any content rendered when modals are closed

jonambas:
Reposting here for documentation
- https://ui.reach.tech/dialog
- https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html

- After opening, those modals immediately switch VO "focus" to somewhere inside the modal, ours remains on the link
- VO "focus" ends at the end of the modal, ours continues to whatever is after the link

nicklemmon:
> * After opening, those modals immediately switch VO "focus" to somewhere inside the modal, ours remains on the link

So I looked back at old documentation, and focusing the modal itself is OK, but it looks like even that might not be happening

So the content element is getting focus when the modal opens, but this isn't ideal - in our app, since the structure of modals is fairly unpredictable, focusing on the modal title would be ideal - but the modal component doesn't come with the title baked in.

Screen Shot 2019-09-24 at 9 05 54 AM

I was wondering what you thought about baking the `<Panel>` UI in to the `<Modal>`? I don't think I have seen a modal that didn't use it at this point, not to mention I have seen differing implementations across the app as a result of how composable it is currently.

> * VO "focus" ends at the end of the modal, ours continues to whatever is after the link

I'm not sure what you mean here? It appears to be working as expected on my end.

PR #252, opened 2019-09-24
Resolve Issue #251 - Update Tag component styles to vertically center content

### LINKED ISSUE #251: The <Tag/> component's content is not vertically centered
Screen Shot 2019-09-23 at 3 20 23 PM

It looks like a few small adjustments would address it:

```css
.Content {
  display: inline-flex;
  align-items: center;
  // and get rid of vertical-align: middle <-- not doing anything outside of a table cell I don't think?
}
```

nicklemmon:
Resolves #251 

## What Changed?
- Updated display of tag `.Content` from `inline-block` to `inline-flex`
- Added some horizontal spacing for child SVG elements

**Before:**
Screen Shot 2019-09-24 at 8 44 57 AM

**After:**
Screen Shot 2019-09-24 at 8 59 48 AM

## How to Test

Look at variants of `<Tag/>` and verify vertical centering:
http://localhost:9001/?selectedKind=Feedback%7CTag&selectedStory=basic%20tag&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

### DIFF
--- a/packages/matchbox/src/components/Tag/Tag.module.scss
+++ b/packages/matchbox/src/components/Tag/Tag.module.scss
@@ -125,8 +125,12 @@
   font-size: font-size(300);
   line-height: rem(22);
 
-  > * {
-    word-spacing: 3px;
+  // This margin is a little funky, however, appears to be the best route to take to add
+  // spacing between icons and other tag content. Ideally, if Tag instances do not have bare
+  // strings inside, then the sibling combinator could be used (i.e., `> * + svg { margin-left: 2px; }`),
+  // however, that's not possible as bare strings do not count as sibling elements.
+  > svg {
+    margin: 0 2px;
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/matchbox/src/components/Tag/Tag.module.scss
```diff
@@ -114,14 +114,24 @@
 }
 
 .Content {
-  display: inline-block;
+  display: inline-flex;
+  align-items: center;
+  justify-content: space-between;
   vertical-align: middle;
```
jonambas: Nice this works.

Couple questions - Do we still need `vertical-align: middle`? Looks like align-items takes precedence. And do we need `space-between` when the `.Tag` container is an inline element with no set width?

nicklemmon: Good questions. Let me play with it some more.

nicklemmon:
Alright @jonambas - let me know what you think about these tweaks.

I also went ahead and removed that global `transition` property that could cause some perf problems. Same fix we added to the buttons a couple of weeks ago.

PR #270, opened 2019-10-29
[4.0.0] Refreshes global CSS, Adds Box and ThemeProvider components

jonambas:
### What Changed
This PR includes a lot of foundational work for the upcoming redesign. Pre-hibana stories are broken (visually), which is fine. The plan is to refactor these one by one.

- Installs `styled-components`, `styled-system`, and a complete set of `design-tokens`
- Maps tokens into styled system's theme via [theme.js](https://github.com/SparkPost/matchbox/pull/270/files#diff-1b2d2a590d7d8af96ecb27c45dca5b45)
- Refactors global CSS to match defaults for the redesign
- Adds a `Box` component
- Adds a `ThemeProvider` component

The `<Box/>` component is meant to be a generic component with access to every styled-system prop. Its goal is to reduce the amount of ad-hoc CSS in the UI, and make custom component styling easier to manage with direct access to design tokens.

#### Helpful links
- [Styled Components](https://www.styled-components.com/docs)
- [Styled System](https://styled-system.com/)
- [Styled System API](https://styled-system.com/api)
- [Styled System Theme Reference](https://styled-system.com/table)
- [SparkPost Design Tokens](https://github.com/SparkPost/design-tokens/blob/1.0.0-beta/dist/index.common.js)

### How To Test or Verify
- Run `npm run clean` and `npm i` to ensure dependancies are updated
- Visit http://localhost:9001/?path=/story/layout-box--styled-box and other stories for the Box component
- Or visit https://codesandbox.io/s/box-test-lyp6i
- Play around with props
- Provide any feedback around the system prop API or token naming conventions

---

<details>
<summary>Screenshots</summary>

Screen Shot 2019-10-31 at 10 35 21 AM
Screen Shot 2019-10-31 at 10 35 14 AM
</details>

<details>
<summary>PR Checklist</summary>

- [x] Get approval from a UX team member (**#uxfe** or **#design-guild** on Slack) for any visual changes
</details>


packages/matchbox/src/styles/global.scss
```diff
@@ -1,90 +1,121 @@
...
+ */
+
+@import '~@sparkpost/design-tokens/tokens.scss';
 @import 'vendor/normalize';
 
 html {
   position: relative;
   box-sizing: border-box;
-  font-size: $base-font-size + 0px;
+  font-size: $DEFAULT_FONT_SIZE + 0px;
```
jonambas: Remove `+ 0px` because this variable already includes `px`

stories/overlays/Modal.stories.js
```diff
@@ -53,12 +53,8 @@ storiesOf('Overlays|Modal', module)
     )
   })
 
-  .add('Toggle Example', withInfo({
```
jonambas: I disabled this initially because there was some conflict while using addon-info with the old story format and new at the same time, but I think I can enable this again

packages/matchbox/src/components/Box/Box.js
```diff
@@ -0,0 +1,47 @@
...
+  ...propTypes.layout,
+  ...propTypes.position,
+  ...propTypes.shadow,
+  ...propTypes.space,
+  ...propTypes.typography,
+  truncate: PropTypes.bool
+};
+
+Box.displayName = 'Box';
+export default Box;
```
nicklemmon: Can arbitrary props be passed in to this component, i.e., data attributes or `aria-label` or things like that?

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * - https://www.styled-components.com/docs/advanced#theming
+ * - https://styled-system.com/theme-specification
+ *
+ * @prop theme Overrides matchbox's theme
+ */
+
+function Theme(props) {
+  return (
+    <ThemeProvider theme={{ ...theme, ...props.theme }}>
```
nicklemmon: :thumbsup: Nice

packages/matchbox/src/components/ThemeProvider/tests/ThemeProvider.test.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import ThemeProvider from '../ThemeProvider';
+
+describe('ThemeProvider', () => {
+  const subject = (props) => shallow(<ThemeProvider {...props} />);
+
+  it('should use the design system theme', () => {
+    expect(subject().prop('theme')).toMatchSnapshot();
```
nicklemmon: Not sure this snapshot offers much of value - it will get out of date when these values change, which will be a pain to manage in the long run.

packages/matchbox/src/components/Box/tests/Box.test.js
```diff
@@ -0,0 +1,28 @@
...
+        borderRadius="pill"
+        border="400"
+        fontSize="200"
+        lineHeight="200"
+        fontFamily="monospace"
+      >
+        Just a Box
+      </Box>
+    );
+    expect(wrapper).toMatchSnapshot();
```
nicklemmon: This snapshot will get out of date easily with these values - it might be worth mocking out the theme config values so that doesn't happen?

packages/matchbox/src/components/ThemeProvider/theme.js
```diff
@@ -0,0 +1,212 @@
...
+    full: '100%'
+  },
+
+  // Example: <Box borderRadius="pill" />
+  radii: {
+    0: tokens.borderRadius_0,
+    100: tokens.borderRadius_100,
+    200: tokens.borderRadius_200,
+    pill: tokens.borderRadius_pill,
+    circle: tokens.borderRadius_circle
```
nicklemmon: Great idea here 👍

packages/matchbox/src/components/Box/Box.js
```diff
@@ -0,0 +1,47 @@
...
+  border,
+  color,
+  flexbox,
+  grid,
+  layout,
+  position,
+  shadow,
+  space,
+  typography
+);
```
nicklemmon: Would there be a separate PR for incorporating these in to existing components/maybe refactoring styling with styled components?

packages/matchbox/src/components/ThemeProvider/tests/ThemeProvider.test.js
jonambas: Yeah good point, I was on the fence about this, but I think a test like this should be in the tokens repo instead of here

packages/matchbox/src/components/Box/Box.js
jonambas: Yep, I'll add a test to make sure valid attributes get passed through

packages/matchbox/src/components/Box/Box.js
jonambas: I'm not 100% sure how much of an effort moving everything to styled-components will be. But for now my plan is to use primitives (`Box` and maybe another one called `Text`) in other components, and see how far that gets us. Most likely we'll be doing a separate PR for each component starting with smaller ones to keep review light

jonambas:
@nicklemmon Think I got everything. Ended up having to mock theme.js globally in the helpers file.

packages/matchbox/src/components/ThemeProvider/theme.js
```diff
@@ -0,0 +1,212 @@
+/* eslint-disable max-lines */
+import tokens from '@sparkpost/design-tokens';
```
gfriedmansp: Is there any way we could have exported constants to use instead of having to depend on all of these string props? Feels like it's pretty susceptible to typos and we'd get free IDE integration for autocomplete. Haven't messed with styled components before, so I'm not sure how much of a pain it would be.

stories/layout/Box.stories.js
```diff
@@ -0,0 +1,45 @@
...
+
+addDecorator((storyFn) => <ThemeProvider>{storyFn()}</ThemeProvider>);
+
+export default {
+  title: 'Layout|Box',
+};
+
+export const StyledBox = withInfo()(() => (
+  <Box
+    color="gray.1000"
```
gfriedmansp: Is there any way we could have exported constants to use instead of having to depend on all of these string props? Feels like it's pretty susceptible to typos and we'd get free IDE integration for autocomplete. Haven't messed with styled components before, so I'm not sure how much of a pain it would be.

jonambas: We can always import tokens directly from `@sparkpost/design-tokens`, but I think it will be fine once we learn token names and when documentation is set up. 

We could in the future explore adding custom proptype definitions or even typescript to help a bit, but one of the benefits of using styled-system is the flexibility from using strings. eg: `color='#000000'` or `width='100px'`, which we wouldn't be able to check for

PR #274, opened 2019-11-08
Upgrade React to 16.11.0

### LINKED ISSUE #273: Update React to the latest version, 16.11.0
Couple points to consider:
- Remove deprecated lifecycle methods
- Remove instances where props not meant for HTML elements are spread via `{...rest}`

Reference: https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html
Audit of spread props: [react-upgrade-to-16-11-0.md.txt](https://github.com/SparkPost/matchbox/files/3852265/react-upgrade-to-16-11-0.md.txt)

nicklemmon:
Resolves #273 

### What Changed
- Upgrades React and React DOM in all packages to 16.11.0
- Prepends deprecated lifecycle methods with `UNSAFE_`
- Replaces `a` links with inline javascript with buttons

### How To Test or Verify
- Open developer tools and verify no React warnings or errors are displayed on all stories
- (Optional) `npm link` with 2web2ui to make sure no additional changes are necessary

### PR Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21068,14 +21068,13 @@
       }
     },
     "react": {
-      "version": "16.8.6",
-      "resolved": "https://registry.npmjs.org/react/-/react-16.8.6.tgz",
-      "integrity": "sha512-pC0uMkhLaHm11ZSJULfOBqV4tIZkx87ZLvbbQYunNixAAvjnC+snJCg0XQXn9VIsttVsbZP/H/ewzgsd5fxKXw==",
+      "version": "16.11.0",
+      "resolved": "https://registry.npmjs.org/react/-/react-16.11.0.tgz",
+      "integrity": "sha512-M5Y8yITaLmU0ynd0r1Yvfq98Rmll6q8AxaEe88c8e7LxO8fZ2cNgmFt0aGAS9wzf1Ao32NKXtCl+/tVVtkxq6g==",
       "requires": {
         "loose-envify": "^1.1.0",
         "object-assign": "^4.1.1",
-        "prop-types": "^15.6.2",
-        "scheduler": "^0.13.6"
+        "prop-types": "^15.6.2"
       },
       "dependencies": {
         "prop-types": {
@@ -21097,11 +21096,6 @@
               }
             }
           }
-        },
-        "react-is": {
-          "version": "16.8.6",
-          "resolved": "https://registry.npmjs.org/react-is/-/react-is-16.8.6.tgz",
-          "integrity": "sha512-aUk3bHfZ2bRSVFFbbeVS4i+lNPZr3/WM5jT2J5omUVV1zzcs1nAaf3l51ctA5FFvCRbhrH0bdAsRRQddFJZPtA=="
         }
       }
     },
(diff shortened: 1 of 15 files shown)


nicklemmon:
It looks like spreading arbitrary props to rendered DOM elements is causing some warnings, one that is easy to reproduce stems from the `ActionList` component.

```js
const {
      actions = [],
      sections = [],
      maxHeight = 'none',
      groupByKey,
      ...rest
    } = this.props;
```

The use of `...rest` means that invalid attributes get added to the DOM like `section` when rendered. Just wanted to make a note of that.

jonambas:
New warning on Snackbars:
```
Warning: A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead. React was passed "javascript:void(0);".
```

jonambas:
> ```js
>       actions = [],
> ```

Looking at the UI, the only props we're spreading into ActionList are `onClick` and `className` which should be fine I think? I think deferring to React's warning might be a good thing here, instead of whitelisting every possible html attr

jonambas:
Slider story is broken, can be fixed by installing `scheduler`

nicklemmon:
I went through the app and (hopefully) found all of the instances of components taking advantage of Matchbox prop spreading. It wouldn't let me upload an `.md` file, but just remove the `.txt` to get syntax highlighting:

[react-upgrade-to-16-11-0.md.txt](https://github.com/SparkPost/matchbox/files/3852265/react-upgrade-to-16-11-0.md.txt)

nicklemmon:
> > It looks like spreading arbitrary props to rendered DOM elements is causing some warnings, one that is easy to reproduce stems from the `ActionList` component.
> > ```js
> > const {
> >       actions = [],
> >       sections = [],
> >       maxHeight = 'none',
> >       groupByKey,
> >       ...rest
> >     } = this.props;
> > ```
> > 
> > 
> > The use of `...rest` means that invalid attributes get added to the DOM like `section` when rendered. Just wanted to make a note of that.
> 
> @nicklemmon Looking at the UI, the only props we're spreading into ActionList are `onClick` and `className` which should be fine I think? I think deferring to React's warning might be a good thing here, instead of whitelisting every possible html attr

@jonambas - apologies for the delayed reply - yeah definitely - I think just exposing those two things is fine for now.

jonambas:
> We could start by removing the spreading and exposing all of these on these particular props?

@nicklemmon I'm not sure this is the right way to go. This could potentially create a lot of work for us in the future. We'd have to cherry pick basically every valid HTML attribute and React `onEvent` handlers.

I propose:
- Cherry picking out the props Matchbox components use directly – props that we know are not valid, or props that we want to explicitly document like `className` and `style`
- Letting React throw errors when an invalid prop is passed through. I think this is fine to put the burden of using the correct props on the developer implementing the components.

nicklemmon:
> > We could start by removing the spreading and exposing all of these on these particular props?
> 
> @nicklemmon I'm not sure this is the right way to go. This could potentially create a lot of work for us in the future. We'd have to cherry pick basically every valid HTML attribute and React `onEvent` handlers.
> 
> I propose:
> 
> * Cherry picking out the props Matchbox components use directly – props that we know are not valid, or props that we want to explicitly document like `className` and `style`
> * Letting React throw errors when an invalid prop is passed through. I think this is fine to put the burden of using the correct props on the developer implementing the components.

Hmm, not sure I follow? Are you saying expose things like `className`, `style`, and `onClick` but not worry about anything else? I was thinking we should remove prop spreading entirely at the Matchbox level is the safest way to go, because otherwise we'll have to update the app in a lot of different places to address this.

nicklemmon:
This is extremely out of date! Closing this.

PR #294, opened 2019-12-04
FE-773 - Add Cypress tests for the Toggle component

nstanard:
FE-773 - Toggle component Cypress tests

### What Changed
Adds Cypress testing to the matchbox toggle component.

### How To Test or Verify

Run storybook locally -> `npm run start:storybook`
Run the e2e tests -> `npm run test:e2e:gui`

### PR Checklist

### DIFF
--- a/cypress/integration/Toggle.spec.js
+++ b/cypress/integration/Toggle.spec.js
@@ -1,0 +1,91 @@
+/// <reference types="Cypress" />
+
+/* eslint-disable no-undef */
+describe('The Toggle component', () => {
+  beforeEach(() => {
+    cy.visit(
+      '/iframe.html?selectedKind=Action%7CToggle&selectedStory=basic%20toggle&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel'
+    );
+
+    // Add data-id attribute
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
+  });
+
+  it('Toggles!', () => {
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('[data-id="toggle-input"]').uncheck({ force: true });
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('[data-id="toggle-input"]').check({ force: true });
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('[data-id="toggle-input"]').uncheck({ force: true });
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+  });
+});
+
+/* eslint-disable no-undef */
+describe('Cannot Toggle component (disabled)', () => {
+  beforeEach(() => {
+    cy.visit(
+      '/iframe.html?selectedKind=Action%7CToggle&selectedStory=disabled%20toggle&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel'
+    );
+
+    // Add data-id attribute
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
+  });
+
+  it('Doesnt Toggle!', () => {
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+  });
+});
+
+/* eslint-disable no-undef */
+describe('Compact Toggle component', () => {
+  beforeEach(() => {
+    cy.visit(
+      '/iframe.html?selectedKind=Action%7CToggle&selectedStory=compact%20toggle&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel'
+    );
+
+    // Add data-id attribute
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
+  });
+
+  it('Toggles!', () => {
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+  });
+});
+
+/* eslint-disable no-undef */
+describe('Compact Toggle component (disabled)', () => {
+  beforeEach(() => {
+    cy.visit(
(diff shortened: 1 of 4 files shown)


cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.get('label > input').should('be.checked');
+    cy.get('label').click();
+    cy.get('label > input').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('label > input').should('be.checked');
+    cy.get('label > input').focus();
+  });
+
+//   it('Focus toggles!', () => {
+//     cy.get('label > input').focus(); // need to do space bar toggle then check
```
nstanard: @nicklemmon - I was trying `keypress` events using `trigger()` as well as `type()` calls but can't get the spacebar event to toggle the checkbox value through Cypress.

nstanard: @nicklemmon <3

nicklemmon:
hey @nstanard - is this ready for review?

nstanard:
@nicklemmon - https://github.com/SparkPost/matchbox/pull/294#issuecomment-564099233 - it is now! 🎊

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+/* eslint-disable no-undef */
+describe('The Toggle component', () => {
+  beforeEach(() => {
+    cy.visit(
+      '/iframe.html?selectedKind=Action%7CToggle&selectedStory=basic%20toggle&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel'
+    );
+
+    // Add data-id attribute
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
```
nicklemmon: This is interesting! Never even thought to do this.

I think it would be OK to add the data attribute in the Storybook file, just to keep things more consistent with other elements. In the app, for example, we do actually render out these attributes so they can interact with Pendo. Obviously we don't need analytics for our Storybook site, but this would just keep things more in line with expectations elsewhere.

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
+/// <reference types="Cypress" />
+
+/* eslint-disable no-undef */
```
nicklemmon: You can add an `.estlinrc` in the `/cypress` folder that disables that rule, that way it doesn't need to be added in each file. I only learned this recently - probably would have included it in the initial Cypress setup otherwise.

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
...
+  it('Toggles!', () => {
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('[data-id="toggle-input"]').uncheck({ force: true });
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('[data-id="toggle-input"]').check({ force: true });
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('[data-id="toggle-input"]').uncheck({ force: true });
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
```
nicklemmon: Is `force` needed? If so, that's fine, might just be worth adding a comment to explain why.

I don't think repeating the toggling this many times is necessary either - just on, off, then on again should probably do it.

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+});
+
+/* eslint-disable no-undef */
+describe('Cannot Toggle component (disabled)', () => {
```
nicklemmon: In general, we try to group our `describe()` statements inside the initial one - this is something I only learned was possible when I started working here:

```js
describe('The toggle component', () => {
  describe('the disabled state', () => {
    // tests go here
  });
});
```

This also makes it a little more DRY so you don't have to repeat your `beforeEach()`!

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
+  });
+
+  it('Doesnt Toggle!', () => {
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
```
nicklemmon: Definitely the right way to test this - just a little repetitive

cypress/integration/Toggle.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    cy.get('label > input').invoke('attr', 'data-id', 'toggle-input');
+  });
+
+  it('Toggles!', () => {
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('not.be.checked');
+    cy.get('label').click();
+    cy.get('[data-id="toggle-input"]').should('be.checked');
```
nicklemmon: Perfect 👍

unreleased.md
```diff
@@ -2,3 +2,4 @@
 
 - #298 - Fix slider tracker rendering with positive min values
 - #298 - Slider now passes through the attribute 'aria-controls'
+- #294 - Add Cypress tests for the Toggle component
```
nicklemmon: No need to add this to this document, actually - probably a little confusing in the PR. Since this isn't a change that actually effects the Matchbox API, no need to list it.

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -56,6 +56,7 @@ class Toggle extends Component {
           onFocus={onFocus}
           onBlur={onBlur}
           type='checkbox'
+          data-id="toggle-input"
```
nicklemmon: We don't want to add it directly to the component, we want to pass it in as a prop. I'll show you in the next file

stories/action/Toggle.stories.js
```diff
@@ -1,14 +1,13 @@
...
-import { Toggle, Checkbox } from '@sparkpost/matchbox';
+import { Toggle } from '@sparkpost/matchbox';
 
 export default storiesOf('Action|Toggle', module)
   .addDecorator((getStory) => (
-    <StoryContainer bg='white'>{ getStory() }</StoryContainer>
+    <StoryContainer bg='white'>{getStory()}</StoryContainer>
   ))
 
   .add('basic toggle', withInfo()(() => (
```
nicklemmon: So in this file, you would do `<Toggle data-id="toggle-input"/>` for the stories you're interested in testing.

So we don't want the `data-id` hardcoded in to every instance of the component, rather, pass it in as needed.

PR #302, opened 2019-12-13
FE-796 List matchbox-icons as a peer dependency

jonambas:
### What Changed
- `matchbox-icons` is now listed as a peer dependency in `matchbox` to reduce bundle size

### How To Test or Verify
(After publishing)
- Install `matchbox` in an app, and verify icons are not bundled with the components
- Can use https://marketplace.visualstudio.com/items?itemName=wix.vscode-import-cost
- Or verify npm warns of a peer dependency when installing, while icons are uninstalled

### PR Checklist


jonambas:
Should this be considered a breaking change?

PR #304, opened 2019-12-17
FE-774 Fix min value default bug & Add Slider cypress tests

jonambas:
### What Changed
- Adds Cypress tests for the Slider component
- Fixes a `min` bug discovered during testing, which did not correctly initially set the sliders value when `min` is greater than `0` on mount

### How To Test or Verify
- Run unit test with `npm run test:unit`
- Run Cypress tests while running storybook with `npm run start:storybook` and `npm run test:e2e:gui` simultaneously

### PR Checklist

### DIFF
--- a/cypress/integration/Slider.spec.js
+++ b/cypress/integration/Slider.spec.js
@@ -17,7 +17,7 @@ describe('The Slider component', () => {
       // Wait until slider has fully rendered with dimensions
       cy.waitUntil(() => cy.get('[data-id="slider-test"]').then(el => {
         return el[0].style.left === '470px';
-      }));
+      }), { timeout: 8000 });
       cy.get('[data-id="slider-wrapper"]').trigger('mousedown', { button: 0, pageX: 200 });
       cy.get('[data-id="slider-wrapper"]').trigger('mouseup');
       cy.get('[data-id="slider-test"]').should('have.attr', 'aria-valuenow', '109')
--- a/cypress/integration/Slider.spec.js
+++ b/cypress/integration/Slider.spec.js
@@ -1,0 +1,82 @@
+/// <reference types="Cypress" />
+
+/* eslint-disable no-undef */
+describe('The Slider component', () => {
+
+  beforeEach(() => {
+    // Setting viewport dimensions to avoid side effects
+    cy.viewport(1000, 660);
+
+    // Need to wait for the slider's ref to mount and calculate dimensions
+    // TODO: Do not calculate handle & track position until ref is available
+    cy.wait(20);
+  });
+
+  describe('when enabled', () => {
+    beforeEach(() => {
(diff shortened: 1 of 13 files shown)


cypress/integration/Slider.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    beforeEach(() => {
+      cy.visit('/iframe.html?selectedKind=Form%7CSlider&selectedStory=basic%20slider');
+    });
+
+    it('should update the sliders value when clicking on the track', () => {
+      cy.wait(200); // Gives slider some time for calculations
+
+      cy.get('[data-id="slider-wrapper"]').trigger('mousedown', { button: 0, pageX: 200 });
+
+      cy.wait(200); // Gives slider some time for calculations
```
jonambas: Ive given up

cypress/integration/Slider.spec.js
```diff
@@ -0,0 +1,83 @@
+/// <reference types="Cypress" />
+
+/* eslint-disable no-undef */
```
nicklemmon: If you drop an extra `.eslintrc` inside the `/cypress` folder you can turn this rule off, but only for JS files in the folder. Check it out:
https://github.com/SparkPost/2web2ui/blob/master/cypress/.eslintrc

(Only recently discovered this was possible!)

cypress/integration/Slider.spec.js
nicklemmon: I don't mind doing this - this is a relatively small library that won't have many tests - if it gets flaky, it can be removed. Maybe add a comment to that effect? Something like `// Remove if introducing flakiness` or something?

packages/matchbox/src/components/Slider/Slider.js
```diff
@@ -12,7 +12,7 @@ function Slider(props) {
   const { defaultValue, disabled, id, max, min, onBlur, onFocus, onChange, precision, ticks, value } = props;
 
   const windowSize = useWindowSize(50);
-  const [sliderValue, setSliderValue] = React.useState(value || defaultValue);
+  const [sliderValue, setSliderValue] = React.useState(value || defaultValue != null ? defaultValue : min);
```
nicklemmon: Would this not work? (instead of using `!=null`). I might be missing something.
```
const [sliderValue, setSliderValue] = React.useState(value || defaultValue ? defaultValue : min);
```

jonambas: So i think I was trying to capture: If defaultValue is set, use it, otherwise use min. This allows `0` to still be valid. The bug was essentially that on first mount, `defaultValue` would sometimes be out of range.

jonambas:
> 👍 Sorry, rapid fire dumbness. You are already testing for the scenario I was concerned about.
> 
> We'll just need to double check the app when this change goes in to make sure it works.

The app will need a snapshot update at the very least because of the defaultProp removal

PR #309, opened 2020-01-06
FE-771 Restyle Button Component

jonambas:
### What Changed
- Restyles the Button and Button.Group components with styled-components
- Matches [these Invision mockups ](https://sparkpost.invisionapp.com/share/FJUVWXX7BXP#/393107481_Button_Symbols)
- Makes use of the deprecate prop helper on:
  - `plain`
  - `Component`
  - `primary`
- Replaces `component` proptype with `elementType` to support strings such as `div` or `a`
- Removes the automatic blur of buttons after clicking, to better support native functionality & accessibility
- Makes updates to Panel section actions, Page primary actions, and Empty State primary actions
- Does NOT add any styled-system props to Button

Screen Shot 2020-01-07 at 10 56 08 AM

Screen Shot 2020-01-07 at 10 59 48 AM

### How To Test or Verify
- Run Storybook with `npm run start:storybook`
- Visit the following stories to verify:
  - Button
  - Page (primary action)
  - Empty State (primary action)
  - Panel (section actions)

### Note
- There is one button style missing from this PR, to be used as connected components to TextFields

### PR Checklist


packages/matchbox/src/components/Panel/tests/Panel.test.js
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 import Panel from '../Panel';
-import { mount } from 'enzyme';
+import { shallow } from 'enzyme';
```
nicklemmon: I think we are actually moving away from shallow mounting on the app team - since the components that make up another component are technically implementation details. What was your thinking here?

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,136 @@
...
+export const base = (props) => `
+    display: inline-block;
+    border-radius: ${tokens.borderRadius_100};
+    user-select: none;
+    font-weight: ${tokens.fontWeight_medium};
+    text-decoration: none;
+    white-space: nowrap;
+    vertical-align: middle;
+    transition-property: background, color, border, outline;
+    transition-duration: .15s;
```
nicklemmon: Should transition duration be a design token? Same with a `transition-timing-function`. I usually feel like `ease-in-out` looks a little nicer than the default (which I believe is `linear`).

packages/matchbox/src/components/Button/Button.js
```diff
@@ -1,140 +1,146 @@
...
-  }
-
-  static defaultProps = {
-    size: 'default'
-  }
+import { base, size, colorVariant, disabled, fullWidth, group } from './styles';
+
+const StyledButton = styled(Box)`
+  ${base}
+  ${size}
```
nicklemmon: This should probably be renamed to avoid conflicts with `styled-system` in the future

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,136 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+export const base = (props) => `
+    display: inline-block;
+    border-radius: ${tokens.borderRadius_100};
+    user-select: none;
+    font-weight: ${tokens.fontWeight_medium};
```
nicklemmon: Love it.

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,136 @@
...
+    transition-property: background, color, border, outline;
+    transition-duration: .15s;
+    border: 1px solid transparent;
+  `;
+
+export const size = (props) => {
+  switch (props.visualSize) {
+    case 'large':
+      return `
+        line-height: calc(3.5rem - 2px);
```
nicklemmon: Is this to handle text wrapping or vertical centering? If the latter, the component should just be styled using `display: inline-flex` and `align-items: center` instead. I could be missing something though.

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  switch (props.visualWeight) {
+    case 'strong':
+      return `
+        &, &:visited {
+          background: ${color};
+          color: ${tokens.color_white};
+    
+          :not([disabled]):hover {
```
nicklemmon: This does increase the CSS specificity considerably, making overrides more difficult. Could the hover disabling be applied to the disabled styles only rather than sort of disabling the disabled hover state here?

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,136 @@
...
+        padding: 0 ${tokens.spacing_400};
+      `;
+  }
+};
+
+export const colorVariant = (props) => {
+  let color;
+  let hoverColor;
+
+  switch (props.buttonColor) {
```
nicklemmon: A great example of how powerful CSS in JS can be

unreleased.md
```diff
@@ -8,4 +8,6 @@
 - #272 - Adds primitive Text component
 - #277 - Removes usage of default token variants
 - #278 - Adds @sparkpost/design-tokens to the Matchbox monorepo
-- #280 - Adds a custom deprecate prop type
\ No newline at end of file
+- #280 - Adds a custom deprecate prop type
+- #309 - Restyles the Button and Button.Group components with styled-components
+- #309 - Buttons no longer automatically blur on click
```
nicklemmon: Very important fix!

.storybook/config.js
```diff
@@ -1,6 +1,7 @@
 import * as storybook from '@storybook/react';
 import { setDefaults } from '@storybook/addon-info';
 import { setOptions } from '@storybook/addon-options';
+import '../packages/matchbox/styles.css';
```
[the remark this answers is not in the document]
jonambas: Let me revert the GH page to the old version and remove this line. For some reason styles weren't loading correctly with a storybook build, but fine when running locally, probably webpack related

packages/matchbox/src/components/Button/Button.js
```diff
@@ -1,140 +1,146 @@
...
-        </WrapperComponent>
-      );
-    }
+  if (to && !WrapperComponent) {
+    return (
+      <StyledButton
+        as='a'
+        href={to}
+        target={external ? '_blank' : ''}
+        rel={external ? 'noopener noreferrer' : ''}
```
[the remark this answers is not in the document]
jonambas: Good idea 👍

packages/matchbox/src/components/Button/tests/Button.test.js
```diff
@@ -47,9 +47,7 @@ describe('Button', () => {
     { name: 'onFocus', event: 'focus' }
   ]);
 
-  it('should handle mouse up', () => {
-    const e = { currentTarget: { blur: jest.fn() }};
-    wrapper.simulate('mouseUp', e);
-    expect(e.currentTarget.blur).toHaveBeenCalled();
+  it('renders base CSS and HTML attributes correctly', () => {
+    expect(global.renderStyled(<Button>Test</Button>)).toMatchSnapshot();
```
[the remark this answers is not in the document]
jonambas: I'm not really sure how to test this. I don't like CSS in snapshots, but doing a full render will print styled-components hashed class names. Let me try to rewrite these tests to assert on specific html attributes.

packages/matchbox/src/components/Button/styles.js
jonambas: Yeah definitely should have a set of tokens for animation – will try to find time to discuss with UX team

packages/matchbox/src/components/Button/styles.js
jonambas: The `-2px` is there to ensure the buttons are a certain height (`3.5rems`) with the `1px` borders. Let me see if I can do this with inline-flex

packages/matchbox/src/components/Button/styles.js
jonambas: Good idea 👍

packages/matchbox/src/components/Button/Button.js
```diff
@@ -1,140 +1,146 @@
...
+const StyledGroup = styled(Group)`
+  ${group(StyledButton)}
+`;
+
+function Button(props) {
+  const {
+    children,
+
+    // Styles
+    primary, // Deprecate in favor of color
```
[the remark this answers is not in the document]
jonambas: Hmm, so I think we could either handle the deprecated colors through wrappers in the app, or I could handle them here, eg. map `orange` -> `blue`, and `navy` `purple` to gray. What do you think?

And yeah, `variant` would be something we could use. Was also thinking about something like `weight` `strong | normal | light`, but we can decide on this later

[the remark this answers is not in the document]
nicklemmon: I think handling them in the app makes sense actually - just theoretically, let's say a second app were to be created and the installed the Hibana version of Matchbox. They would just expect that purple would be purple and orange would be orange, so then no need to handle it in the component itself. 👍

packages/matchbox/src/components/Button/Button.js
```diff
@@ -1,140 +1,147 @@
...
-      );
-    }
+  if (to && !WrapperComponent) {
+    return (
+      <StyledButton
+        as='a'
+        href={to}
+        target={external ? '_blank' : ''}
+        rel={external ? 'noopener noreferrer' : ''}
+        title={external ? 'Opens in a new tab' : ''}
```
nicklemmon: It looks like this renders an empty `title` attribute when the link is not `external`.

Maybe we could have a `title` prop that would typically be undefined. That would handle the aforementioned empty title problem *and* allow some flexibility:

```
const { title } = props;

<Button title={external ? 'Opens in a new tab' : title}/>
```

packages/matchbox/src/components/Button/Button.js
```diff
@@ -1,140 +1,149 @@
...
+  const WrapperComponent = component || Component;
+
+  // Polyfills to be deprecrated 'primary' and 'destructive' prop
+  const buttonColor = primary ? 'blue' : destructive ? 'red' : color;
+
+  // Experimenting with a weight prop to replace outline, plain, and flat in the future
+  const visualWeight = React.useMemo(() => {
+    if (outline) {
+      return 'normal';
+    }
```
nicklemmon: For some reason the `"outline"` variant is rendering with a gray background in Chrome in responsive mode (so probably on phones it would look like this):

Screen Shot 2020-01-21 at 12 25 18 PM

nicklemmon: Oddly enough, the same thing is happening in Safari with no special settings or anything:
Screen Shot 2020-01-21 at 12 28 53 PM

nicklemmon: *and* Firefox. Weird stuff.

Screen Shot 2020-01-21 at 12 29 52 PM

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,137 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+export const base = (props) => `
+    display: inline-flex;
+    align-items: center;
+    border-radius: ${tokens.borderRadius_100};
+    user-select: none;
```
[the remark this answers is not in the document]
jonambas: This is a good suggestion for a new component.  What do you think about exposing `width` and `margin` styled-system props on Button? Could do things like `width={['100%', 'auto']}`, same with margin.

And when we see a pattern on how they are used we could turn them into wrappers?

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,137 @@
...
+    border-radius: 0;
+    margin-right: -1px;
+
+    &:first-child {
+      border-top-left-radius: ${tokens.borderRadius_100};
+      border-bottom-left-radius: ${tokens.borderRadius_100};
+    }
+    &:last-child {
+      border-top-right-radius: ${tokens.borderRadius_100};
+      border-bottom-right-radius: ${tokens.borderRadius_100};
```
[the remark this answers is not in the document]
jonambas: Yeah I agree, too small to notice with the -1px margin

packages/matchbox/src/components/Button/styles.js
```diff
@@ -0,0 +1,137 @@
...
+export const colorVariant = (props) => {
+  let color;
+  let hoverColor;
+
+  switch (props.buttonColor) {
+    case 'orange': // To be deprecated
+    case 'blue':
+      color = tokens.color_blue_700;
+      hoverColor = tokens.color_blue_800;
+      break;
```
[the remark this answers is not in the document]
jonambas: Thing about design tokens is that we can change colors later 😎

jira ticket: https://jira.int.messagesystems.com/browse/FE-844
asana task: https://app.asana.com/0/1143076153442625/1158329596760225

packages/matchbox/src/components/Button/styles.js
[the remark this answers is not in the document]
nicklemmon: That's a great idea. I like it.

PR #312, opened 2020-01-13
Resolves #301 - Adds Prettier support

### LINKED ISSUE #301: Add Prettier support
Prettier support would help improve the developer experience for this project. Ideally, Prettier could be configured such that:

1. A developer is able to locally enable formatting on save
2. Staged/changed files on a commit are formatted via a git precommit hook

See the setup in 2web2ui for an example setup.

nicklemmon:
Resolves #301 

### What Changed
- Adds Prettier support
- Tests new plugin by making an arbitrary change, then running Prettier both in the precommit hook and on save

### How To Test or Verify
- Checkout the `incorporate-prettier` branch, go to a file, make an arbitrary change
- On commit, the precommit hook should format the file
- If using VSCode, install the `prettier-vscode` extension (https://github.com/prettier/prettier-vscode). On save, Prettier should now format files

### PR Checklist
N/A


.prettierignore
```diff
@@ -0,0 +1 @@
+node_modules
```
jonambas: Do we need anything else in here? Like any of the package built files, package node modules, storybook build etc?

nicklemmon: Hmm, a good thought. Let me take a look in the AM.

nicklemmon: You know, now that I'm looking at it, we don't need this file at all with our current setup since they are all git ignored. Since our set up is:

1. Format on save
2. Format on precommit hook

Any files that are never edited or ignored by git won't be formatted with Prettier.

PR #323, opened 2020-01-30
FE-812 Update ProgressBar Component

logansparlin:

### What Changed

- Updated the ProgressBar component to match the new design
- Set color to always be UI blue and add deprecation message in console

### How To Test or Verify
- `npm run start:storybook`
- compare design changes to https://sparkpost.invisionapp.com/d/main#/console/18874652/395966961/inspect

### PR Checklist


packages/matchbox/src/components/ProgressBar/styles.js
```diff
@@ -0,0 +1,40 @@
...
+export const outerBase = () => `
+  position: relative;
+  background-color: ${tokens.color_gray_300};
+  border-radius: ${tokens.borderRadius_100};
+  width: 100%;
+  margin-bottom: 0.75rem;
+`;
+
+export const innerBase = () => `
+  width: 100px;
```
nicklemmon: I'm not sure why this is necessary? There may be a reason - if so, would you mind writing a comment describing your reasoning here?

packages/matchbox/src/components/ProgressBar/styles.js
```diff
@@ -0,0 +1,40 @@
...
+  background-color: ${tokens.color_blue_700};
+  border-radius: ${tokens.borderRadius_100} ${tokens.borderRadius_0} ${tokens.borderRadius_0} ${tokens.borderRadius_100}
```
nicklemmon: The `100` value for the border radius is very slight - almost unnoticeable in this situation. Might be worth trying a larger value? Totally subjective though.

packages/matchbox/src/components/ProgressBar/tests/ProgressBar.test.js
```diff
@@ -1,26 +1,59 @@
...
-    wrapper.setProps({ completed: 50, color: 'navy' });
-    expect(wrapper).toMatchSnapshot();
-  });
+describe('renders progress bar inner styles', () => {
+  const progressBar = props => global.mountStyled(<ProgressBar {...props} />);
+
+  it('should render progress bar width based on completed percent', () => {
+    const wrapper = progressBar({ completed: 24 });
 
-  it('renders at 100% progress if given a value over 100', () => {
```
nicklemmon: It would be worth leaving a test in for the `completed` value exceeding `100` and going below `1` (maybe even a negative value) just to make sure we are keeping track of that requirement that the bar can't go below 0 or above 100.

stories/feedback/ProgressBar.stories.js
```diff
@@ -1,15 +1,18 @@
...
-  ))
-  .add('Default', withInfo()(() => (
-    <ProgressBar completed={54} color='red' />
-  )));
+  .addDecorator(getStory => <StoryContainer>{getStory()}</StoryContainer>)
+  .add(
+    'Default',
+    withInfo()(() => (
+      <div>
+        <ProgressBar completed={54} color="purple" />
```
nicklemmon: the `color` prop is deprecated, no? Might be confusing leaving it in from a documentation point of view.

packages/matchbox/src/components/ProgressBar/styles.js
```diff
@@ -0,0 +1,40 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+export const outerBase = () => `
+  position: relative;
+  background-color: ${tokens.color_gray_300};
+  border-radius: ${tokens.borderRadius_100};
+  width: 100%;
+  margin-bottom: 0.75rem;
```
nicklemmon: I don't think including a baked-in `margin-bottom` is a good idea - rather, developers should just use the exposed `styled-system` props to position each component instance. You can do the same in the Storybook file!

packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,45 +1,51 @@
...
+  color: deprecate(
+    PropTypes.oneOf(['orange', 'blue', 'navy', 'purple', 'red']),
+    'Color is always blue for now. This may be updated in the future.',
+  ),
+  size: PropTypes.oneOf(['normal', 'small']),
+  ...createPropTypes(margin.propNames),
+};
+
+ProgressBar.defaultProps = {
+  size: 'normal',
```
jonambas: I think default props changes still surface in Enzyme snapshots, so this would be a slightly-breaking change for the app. I think I would destructure `normal` to a default value in the component instead, like how `completed` is now

actually it might not matter at all since we're changing the color and completed default props too /shrug

packages/matchbox/src/components/ProgressBar/styles.js
logansparlin: Not necessary, deleting

packages/matchbox/src/components/ProgressBar/styles.js
logansparlin: I'm gonna leave as is since it's specified in the design files this way

stories/feedback/ProgressBar.stories.js
```diff
@@ -1,15 +1,18 @@
...
 import { ProgressBar } from '@sparkpost/matchbox';
 
 storiesOf('Feedback|ProgressBar', module)
-  .addDecorator((getStory) => (
-    <StoryContainer>{ getStory() }</StoryContainer>
-  ))
-  .add('Default', withInfo()(() => (
-    <ProgressBar completed={54} color='red' />
-  )));
+  .addDecorator(getStory => <StoryContainer>{getStory()}</StoryContainer>)
```
jonambas: In order to get system props working with the tokens the stories need to be wrapped with `ThemeProvider`. Would be nice to refactor this a bit into the new storybook format using `export`s instead, see the Button stories for an example

packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,45 +1,51 @@
...
+  ${innerBase}
+  ${visualSize}
+  ${calculatedWidth}
+`;
+
+function ProgressBar(props) {
+  const { completed = 0, size } = props;
+
+  return (
+    <StyledProgressBarOuter as="div" size={size}>
```
jonambas: Will need to spread props here to get any of the margin props passed through 
```js
const { completed = 0, size, ...rest } = props;
<StyledProgressBarOuter as="div" size={size} {...rest}>
```

jonambas: Spreading props is a bad pattern though that we'll have to fix eventually but I'll create an issue to fix this sort of thing holistically, at least for system props

packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,45 +1,51 @@
...
+      <StyledProgressBarInner as="div" completed={completed} size={size} />
```
jonambas: Small bug here - `size` as a prop when passed through Box gets translated into `width` and `height`, would be worth renaming to `visualSize` to match that style function above ie `visualSize={size}`

jonambas:
Sorry, this comment is coming a little late. Do we want to pull in any of the changes made in #245? @nicklemmon thoughts?

nicklemmon:
> Sorry, this comment is coming a little late. Do we want to pull in any of the changes made in #245? @nicklemmon thoughts?

@jonambas I think not yet, actually - when we get to the point where we are changing component APIs we should bring this stuff in. Looking at that issue, we'll need some new props to handle things appropriately!

packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -1,45 +1,51 @@
...
-     */
-    completed: PropTypes.number.isRequired,
-
-    /**
-     * Bar color
-     */
-    color: PropTypes.oneOf(['orange', 'blue', 'navy', 'purple', 'red'])
-  };
-
-  static defaultProps = {
```
jonambas: Just so we don't forget later, could we make a note of the default prop changes in `unreleased.md`?

PR #329, opened 2020-02-11
FE-807 Update Panel Component

logansparlin:

### What Changed

- Updated the Panel component to match new design
- Color defaults to blue 

### How To Test or Verify
npm run start:storybook
compare design changes to https://sparkpost.invisionapp.com/d/main#/console/18874652/397154347/preview

### PR Checklist


jonambas:
Tests are failing here because coverage dropped, which is set in the jest config:
```
"coverageThreshold": {
    "global": {
      "statements": 90
    }
  },
```

There are a few lines here you can maybe test: https://travis-ci.org/SparkPost/matchbox/builds/648771857#L589
But don't sweat if it seems like overkill, we can lower that number too

packages/matchbox/src/components/Panel/Accent.js
```diff
@@ -0,0 +1,20 @@
...
+  ${accent}
+`;
+
+const Accent = ({ accentColor }) => {
+  return <AccentOuter accentColor={accentColor} />;
+};
+
+Accent.displayName = 'Panel.Accent';
+Accent.propTypes = {
+  accentColor: PropTypes.oneOf(['orange', 'blue', 'red', 'yellow', 'green', 'gray']),
```
jonambas: Missing `navy`, `purple` from the old props

packages/matchbox/src/components/Panel/Footer.js
```diff
@@ -1,44 +1,48 @@
...
-              {right}
-            </div>
-          </Grid.Column>
-        </Grid>
-      </div>
-    );
-  }
+  return (
+    <OuterWrapper {...rest}>
+      <Grid>
```
jonambas: Let's not use `Grid` here, I think we're going to be deprecating this component pretty soon. You can achieve the same thing with `Box` and some flex properties

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
...
+`;
+
+export const panelInner = props => {
+  return `
+    border: ${tokens.borderWidth_100} solid ${tokens.color_gray_400};
+    border-top-width: ${props.accent ? '0' : '1px'};
+  `;
+};
+
+export const left = () => `
```
jonambas: Not sure we need these styles here, can use `<Box textAlign='left'>`. I think in general we should only use styled-components if we need dynamic styles

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
...
+  margin: 0;
+
+  text-rendering: optimizeLegibility;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+`;
+
+export const panel = () => `
+  position: relative;
+  margin-bottom: ${tokens.spacing_400};
```
jonambas: I think we're trying to remove all outer margins from components, and we can use the `margin` `system-prop` instead when needed

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
...
+  background: ${tokens.color_white};
+`;
```
jonambas: I might be wrong but I think there is a slight border-radius on panels, would need to do the same on the accent bar too

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
...
+    case 'gray':
+      color = tokens.color_gray_600;
+      break;
+    case 'blue':
+    default:
+      color = tokens.color_blue_700;
+  }
+
+  return `
+    height: 3px;
```
jonambas: I think this is 4px, or the smallest spacing token

packages/matchbox/src/components/Panel/Panel.js
```diff
@@ -1,82 +1,87 @@
...
+Panel.Header = Header;
+Panel.Section = Section;
+Panel.Footer = Footer;
+Panel.Accent = Accent;
+
+Panel.propTypes = {
+  title: PropTypes.node,
+  accent: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.oneOf(['orange', 'blue', 'red', 'yellow', 'green', 'gray']),
```
jonambas: Same here, missing a couple colors

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
...
+    background-color: ${color};
+  `;
+};
+
+export const body = () => `
+  display: flex;
+  padding: 18px ${tokens.spacing_500};
```
jonambas: I think this padding needs to be derived from the system-prop, and be defaulted to the 24px token, `500`. Lemme know if this doesn't makes sense

packages/matchbox/src/components/Panel/styles.js
```diff
@@ -0,0 +1,97 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+export const header = () => `
+  display: flex;
+  font-weight: ${tokens.fontWeight_semibold};
+  padding: ${tokens.spacing_500} ${tokens.spacing_500} 0 ${tokens.spacing_500};
```
jonambas: Same here this padding should probably be defined by system-props

packages/matchbox/src/components/Panel/Header.js
```diff
@@ -1,22 +1,46 @@
...
 
 const actionOverrides = { flat: true, size: 'small' };
 
-const Header = ({ title, actions }) => {
-  const actionMarkup = actions && actions.length
-    ? <div className={styles.Actions}>{buttonsFrom(actions, actionOverrides)}</div>
-    : null;
+const Header = ({ title, actions, ...rest }) => {
+  const actionMarkup =
+    actions && actions.length ? <Actions>{buttonsFrom(actions, actionOverrides)}</Actions> : null;
```
jonambas: This is actually a really good place to use the new `Inline` component, same with the actions inside sections

jonambas: Actually, nevermind, this produces a button group, not buttons

PR #334, opened 2020-02-18
FE-794 Update Banner Component

jonambas:
Mockups: https://sparkpost.invisionapp.com/d/main#/console/18874652/395021232/preview

### What Changed
- Restyles the Banner Component
- Dismiss button is now a button and responds to space and enter keydown events, and is now screen reader accessible
- Status 'default' now defaults to 'info', gray banners no longer exist

Screen Shot 2020-02-18 at 6 31 19 PM

### How To Test or Verify
- Visit the Banner stories and verify the component matches the mocks and behaves as expected

### PR Checklist


packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,149 @@
...
-      </div>
-    );
-  }
+    return result;
+  }, [action, actions]);
+
+  const dismissMarkup = onDismiss ? (
+    <Box flex="0">
+      <StyledDismiss as="button" onClick={onDismiss} status={status} color="gray.800">
+        <Close size={24} label="Dismiss" />
```
nicklemmon: I think it would be better to use `<ScreenReaderOnly/>` here and just not have a label for the dismiss button. From what I remember correctly, `aria-label`s inside buttons aren't always encountered by screen readers.

jonambas: Good to know!

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,149 @@
...
+  ${dismissColor}
+`;
+
+function Banner(props) {
+  const { children, title, status, action, actions, onDismiss, ...rest } = props;
+
+  const titleMarkup = title ? (
+    <Text fontSize="500" lineHeight="500" as="h5">
+      {title}
+    </Text>
```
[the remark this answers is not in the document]
jonambas: Good catch, the current banners vertically align the icon with those two lines of text

packages/matchbox/src/components/Banner/styles.js
```diff
@@ -0,0 +1,94 @@
+import { tokens } from '@sparkpost/design-tokens';
+import { ErrorOutline, CheckCircleOutline, InfoOutline } from '@sparkpost/matchbox-icons';
+
```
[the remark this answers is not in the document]
jonambas: Moving to blue 800 for links would work here, although I'm not sure how we can enforce this in this component

jonambas:
I removed the button override styling from the component figured we would need to be more flexible around whats outlined

packages/matchbox/src/components/Banner/styles.js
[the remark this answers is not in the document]
nicklemmon: Hmm, is it possible in styled components to target child elements? Scoping might prevent this.

Basically, is there any way to do this? I know it's not the prettiest CSS in the world, but might be simpler than the alternative:
```
.Banner {
  a {
     color: blue;
  }
}
```

packages/matchbox/src/components/Banner/styles.js
```diff
@@ -0,0 +1,94 @@
...
+    case 'default':
+    default:
+      return `
+        background: ${tokens.color_blue_100};
+        border: 1px solid ${tokens.color_blue_600};
+      `;
+  }
+}
+
+export const statusIcons = {
```
nicklemmon: Looks like some funky things start to happen on smaller viewports:
Screen Shot 2020-02-20 at 9 20 27 AM

packages/matchbox/src/styles/helpers.js
```diff
@@ -0,0 +1,6 @@
+export const buttonReset = `
```
nicklemmon: 👍 Nice.

nicklemmon: You _might_ be missing a few things? Especially for mobile devices:
https://gist.github.com/MoOx/9137295

I am trying to get Xcode installed so I can at least use iOS simulator to test things out. No means of using Android for this sort of thing sadly.

stories/feedback/Banner.stories.js
```diff
@@ -1,66 +1,95 @@
...
+    status="info"
+    title="Verify your email address"
+    action={{
+      content: 'Resend Email',
+      onClick: action('Resend Email Clicked'),
+    }}
+  >
+    Please click the link in the email we sent you to unlock the higher daily sending limits.
+  </Banner>
+));
```
nicklemmon: In these examples, I think ideally the child element would be a `<p>` to be semantically correct - that will encourage future devs to do the same instead of just a string.

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,147 @@
...
+  }, [action, actions]);
+
+  const dismissMarkup = onDismiss ? (
+    <Box flex="0">
+      <StyledDismiss as="button" onClick={onDismiss} status={status} color="gray.800">
+        <ScreenReaderOnly>Dismiss</ScreenReaderOnly>
+        <Close size={24} />
+      </StyledDismiss>
+    </Box>
+  ) : null;
```
nicklemmon: In the future, I wonder if we could use components for this sort of thing instead of variables? This is really not a substantive critique. In general FE devs use variables and functions within the scope of a component for this sort of thing, when I feel like just adding additional (but not exported) components might make more sense and make things more "React-y"? Worth a convo, but not worth changing anything at this point.

packages/matchbox/src/components/Banner/styles.js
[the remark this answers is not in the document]
jonambas: Yea we can do this for sure, but would still have to double check when we use Banners. Depends on what types of links / components are placed in the Banner. 👍 I'll add this rule in

packages/matchbox/src/components/Banner/Banner.js
jonambas: Yup agree 100%

packages/matchbox/src/components/Banner/styles.js
jonambas: This ones a tough example because the button text is causing the overflow, I'll update the story though

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,151 @@
...
 
   return (
-    <div className={styles.IconWrapper}>
-      <Icon size={30} className={iconClasses} label={iconLabel}/>
-      <div className={styles.IconBackdrop} />
-    </div>
+    <Box
+      position="relative"
+      // IconSection do not render under the first breakpoint
+      display={['none', 'none', 'flex']}
```
nicklemmon: That works well. `<Box/>` to the rescue.

The only issue is actually an a11y problem with doing that...technically a user with color blindness won't be able to see what the status of a `<Banner/>` is without the icon there as well on all viewports.

Any other possible routes to take? I have seen some design systems move the icon above the rest of the content, or possibly make it substantially smaller on more narrow viewports.

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,151 @@
...
+      <StyledDismiss as="button" onClick={onDismiss} status={status} color="gray.800">
+        <ScreenReaderOnly>Dismiss</ScreenReaderOnly>
+        <Close size={24} />
+      </StyledDismiss>
+    </Box>
+  ) : null;
+
+  return (
+    <StyledContainer display="flex" p="500" borderRadius="100" status={status} {...rest}>
+      <IconSection status={status} />
```
nicklemmon: 🎉

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,151 @@
...
+    </Box>
+  ) : null;
+
+  const actionMarkup = React.useMemo(() => {
+    let result = action ? <div>{buttonFrom(action)}</div> : null;
 
     if (actions) {
-      actionMarkup = (
-        <div className={styles.Actions}>
-          {actions.map((action, i) => buttonFrom(action, actionOverrides, i))}
```
nicklemmon: We don't need `actionOverrides` anymore?

packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -1,144 +1,151 @@
...
-      </div>
-    );
-  }
+    return result;
+  }, [action, actions]);
+
+  const dismissMarkup = onDismiss ? (
+    <Box flex="0">
+      <StyledDismiss as="button" onClick={onDismiss} status={status} color="gray.800">
+        <ScreenReaderOnly>Dismiss</ScreenReaderOnly>
```
nicklemmon: Nice.

packages/matchbox/src/components/Banner/styles.js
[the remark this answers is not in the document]
nicklemmon: Looks like that fixed it. Nice.

packages/matchbox/src/components/Banner/Banner.js
jonambas: Oh right, good call. I'll probably add the icon back in and align it to the top left of the Banner

packages/matchbox/src/components/Banner/Banner.js
jonambas: The new banners use both outlined / non outlined, normal sized buttons, I figured it would be best to have the flexibility here instead of specifying which buttons should be outlined / non outlined

PR #345, opened 2020-03-06
FE-921 Tag contents and close button are not vertically aligned

bkemper:
This is follow-up to #284 

### What Changed
- Wrapped ComboBoxTextField items with `<div />`'s to control the spacing

Screen Shot 2020-03-05 at 11 28 28 PM

Screen Shot 2020-03-05 at 11 28 17 PM

Screen Shot 2020-03-05 at 11 31 44 PM

Screen Shot 2020-03-05 at 11 32 30 PM

### How To Test or Verify
- Make sure to change the with of the available space to test wrapping


bkemper:
Good catch Jon.  Changing the `$base-font-size` will break the Tag alignment.  To fix it, I remove `align-items: center;` from the Tag container.  This will allow the content and the close button to fill the full vertical space.  Then used `flex` to vertically center the close button svg.

PR #346, opened 2020-03-06
FE-827 - Update Tooltip Component

jonambas:
### What Changed
- Restyles the Tooltip component
- Deprecates the dark prop
- Horizontal offset is ignored for now, will do arrow positioning potentially in the future
- Makes system props available:  border, shadow, background, padding, typography

Screen Shot 2020-03-06 at 4 00 03 PM
Screen Shot 2020-03-06 at 3 59 52 PM

### How To Test or Verify
- Visit Tooltip stories

### PR Checklist


packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -1,124 +1,166 @@
...
-    horizontalOffset: '0px',
-    forcePosition: false,
-    preferredDirection: {
-      bottom: true,
-      left: false,
-      right: true,
-      top: false
-    }
-  }
+const StyledContent = styled(Box)`
```
jonambas: Need to change this – using `Box` allows *all* system props, should use a `'div'` instead

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -1,124 +1,166 @@
...
+      >
+        {props.children}
+      </Box>
     );
   }
+
+  return (
+    <TooltipOverlay
+      eventDebounce={props.eventDebounce}
+      portalId={props.portalId}
```
bkemper: I thought there is an accessibility need to render the tooltip DOM with/alongside the button?

Screen Shot 2020-03-18 at 11 15 51 AM

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -1,124 +1,166 @@
...
+  horizontalOffset: PropTypes.string, // TODO Either deprecate or style Tooltip tips
+  /**
+   * Disables automatic positioning
+   */
+  forcePosition: PropTypes.bool,
+  children: PropTypes.node,
+  /**
+   * These props are provided automatically through the Overlay component
+   */
+  preferredDirection: PropTypes.shape({
```
bkemper: follow-up: Shouldn't this be an array?

packages/matchbox/src/components/Tooltip/TooltipOverlay.js
```diff
@@ -1,76 +1,75 @@
...
-  handleMeasurement = () => {
-    this.setState({
-      preferredDirection: getPreferredDirectionFor(this.activator),
-      position: getPositionFor(this.activator)
-    });
-  }
+  return (
+    <>
+      <WindowEvent event="resize" handler={debounce(handleMeasurement, eventDebounce)} />
+      <WindowEvent event="scroll" handler={debounce(handleMeasurement, eventDebounce)} />
```
bkemper: I know this isn't new, but do we need to call handleMeasurement on resize and scroll events?  The side effect already handles mouseover and mouseout.

packages/matchbox/src/components/Tooltip/Tooltip.js
jonambas: Potentially yeah, also these props don't really need to be defined here and only `top` and `left` are used. Can revise this a bit when we make API updates

packages/matchbox/src/components/Tooltip/TooltipOverlay.js
jonambas: Resize can be removed, but now that we are using fixed positioning, this is the only thing that will hide tooltips if you one is visible before you start scrolling

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -139,15 +141,14 @@ Tooltip.propTypes = {
     left: PropTypes.bool,
     right: PropTypes.bool,
   }),
-  /**
-   * Element ID for the portal that will house tooltips. Appends to body if not provided.
-   */
+
   portalId: PropTypes.string,
```
bkemper: This should probably be removed.

PR #348, opened 2020-03-06
FE-822 Update Tabs Component

jonambas:
### What Changed
- Restyles the Tabs component
- Tabs are now 'button's and are focusable

Screen Shot 2020-03-07 at 10 00 37 AM
Screen Shot 2020-03-07 at 10 00 41 AM

### How To Test or Verify
- Visit Tab stories
  - Verify design matches the mocks
  - Verify functionality

### PR Checklist


packages/matchbox/src/components/Tabs/styles.js
```diff
@@ -0,0 +1,36 @@
+import { buttonReset } from '../../styles/helpers';
+import { tokens } from '@sparkpost/design-tokens';
+
+export const tabStyles = ({ selected, fitted }) => `
+  ${buttonReset}
+  position: relative;
+  flex: ${fitted ? '1' : '0'};
```
bkemper: 👍

packages/matchbox/src/components/Tabs/Tabs.js
```diff
@@ -77,26 +55,45 @@ class Tabs extends Component {
...
+    'Tab color is no longer configurable',
+  ),
+  /**
+   * Index of selected tab
+   */
+  selected: PropTypes.number.isRequired,
+  /**
+   * Connects this component with component underneath it. Works well with Panels.
+   */
+  connectBelow: deprecate(PropTypes.bool, 'Deprecated in favor of margin system props'),
```
bkemper: It looks like this style was removed.  If so, this should be removed or if you want to provide a nice DX, you could provide an error if it is being used.

jonambas: Thats what this `deprecate` function does 😄

bkemper: `deprecate` logs a warning.  That is what I would expect from a prop that is planned to be removed in a future release.  However, the supporting code for this props has been removed.  It would be inaccurate to log a warning.  Either this prop should be removed or it should throw an error to immediately alert the dev that it has been removed.

jonambas: Not 100% sure how to handle that – it's been replaced with system props. Would need to throw an error if the prop is _not_ used. But we're doing the same with almost every other component, so I think it can maybe be announced elsewhere.

PR #349, opened 2020-03-08
FE-818 Update TextField

jonambas:
### What Changed
- Restyles TextField
- Creates a new hook `useInputDescribedBy` used in TextField and Select
- TextField with labelHidden now properly adds a screen reader accessible label
- Outlined Button does not match input styles – will create a new button style in a new ticket: FE-925

Screen Shot 2020-03-09 at 8 45 25 AM

### How To Test or Verify
- Visit TextField stories

### PR Checklist


packages/matchbox/src/helpers/tests/geometry.test.js
```diff
@@ -86,3 +98,13 @@ describe('linear interpolation', () => {
     expect(lerp(-10, 30, 1.5)).toBe(30);
   });
 });
+
+describe('round to baseline', () => {
```
jonambas: Added this, rest is prettier

packages/matchbox/src/helpers/geometry.js
```diff
@@ -99,3 +98,10 @@ export function lerp(min, max, n) {
   const value = (max - min) * n + min;
   return clamp(value, min, max);
 }
+
+/**
+ * Rounds a number to the nearest baseline ceiling
+ */
+export function roundToBaseline(n, base = 4) {
+  return Math.ceil(n / base) * base;
+}
```
jonambas: Added this, rest is prettier

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+
+  const componentProps = omit(rest);
+  const systemProps = pick(rest);
+  const prefixRef = React.useRef(null);
+  const suffixRef = React.useRef(null);
+  const [prefixPadding, setPrefixPadding] = React.useState('400');
+  const [suffixPadding, setSuffixPadding] = React.useState('400');
+  const { describedBy, errorId, helpTextId } = useInputDescribedBy({ id, helpText, error });
+
+  React.useLayoutEffect(() => {
```
nicklemmon: This is really interesting!

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+    if (prefixRef.current) {
+      setPrefixPadding(roundToBaseline(getRectFor(prefixRef.current).width + 16));
```
nicklemmon: Should these `16` values be coming from design tokens?

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+    required,
+    resize = 'both',
+    type,
+    suffix,
+    suffixClassname,
+    style,
+    ...rest
+  } = props;
+
+  const componentProps = omit(rest);
```
nicklemmon: Will this not potentially introduce bugs? We take advantage of prop spreading in the app a good bit.

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
-      </fieldset>
-    );
-  }
+      <Connect left={connectLeft} right={connectRight}>
+        <Box position="relative">
+          <PrefixOrSuffix
+            content={prefix}
+            className={prefixClassname}
+            forwardedRef={prefixRef}
+            left="300"
```
nicklemmon: When focusing a field with a prefix icon, the icon disappears. Is that expected?

Screen Shot 2020-03-11 at 12 26 28 PM

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+          <PrefixOrSuffix
+            content={suffix}
+            className={suffixClassname}
+            forwardedRef={suffixRef}
+            right="300"
+          />
+        </Box>
+      </Connect>
+      {error && !errorInLabel && <Error id={errorId} error={error} />}
+      {helpText ? <HelpText id={helpTextId}>{helpText}</HelpText> : null}
```
nicklemmon: I think you could just do `{helpText && <HelpText/>}` rather than the ternary here

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+const StyledWrapper = styled('div')`
+  ${margin}
+`;
+
+// TODO Make this reusable and shared with Select
+const FieldBox = props => {
+  return (
+    <Box
+      as="input"
+      disabled={props.disabled}
```
nicklemmon: Just like the `<Select/>`, it would be good to use the `required` attribute when the `required` prop is true. Appears to be missing at the moment:

Screen Shot 2020-03-11 at 12 32 53 PM

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+      width="100%"
+      border={props.hasError ? `1px solid ${tokens.color_red_700}` : '400'}
+      borderRadius="100"
+      bg={props.disabled ? 'gray.200' : 'white'}
```
nicklemmon: Maybe adding the `not-allowed` cursor would be a good idea when disabled?

stories/form/TextField.stories.js
```diff
@@ -1,102 +1,85 @@
 import React from 'react';
```
nicklemmon: Some of these stories don't have `id` props passed in - I thought there was a bug with the implementation when Voiceover wasn't working, but then I realized it as just the component instances that were the problem.

packages/matchbox/src/hooks/useInputDescribedBy.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+/**
+ * Generates an aria-describedby attribute and ids for HelpText and Error if id is present
+ *
+ * @example
+ *  const { describedBy, errorId, helpTextId } = useInputDescribedBy({
+ *    id: 'input-id',
+ *    helpText: 'This is helptext',
+ *    error: 'This is an error'
```
nicklemmon: It doesn't look like the actual help text is used in the hook? Just that the help text exists? Could these be booleans like `hasHelpText` and `hasErrorText`?

packages/matchbox/src/components/TextField/TextField.js
nicklemmon: Looks like it's just a missing `id` prop in the Storybook story.

Technically the `id` prop should be required - what would you think about making it required? Could force devs to pass in one for each component instance, improving a11y on the app pretty tremendously fairly easily to boot.

packages/matchbox/src/components/Connect/Connect.js
```diff
@@ -1,46 +1,44 @@
...
-
-    return (
-      <div className={styles.Connect}>
-        {leftMarkup}
-        <div className={primaryClasses}>{children}</div>
-        {rightMarkup}
-      </div>
-    );
+const FocusHandler = styled(Box)`
+  *:focus {
```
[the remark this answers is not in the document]
jonambas: Hmm `*` actually selects everything, haha. This might be a good opportunity to try `focus-within`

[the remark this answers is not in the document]
nicklemmon: Yeah support is good enough for us to use it!

packages/matchbox/src/components/TextField/TextField.js
jonambas: Fixed when using `&:focus-within` instead of `*:focus`

packages/matchbox/src/hooks/useInputDescribedBy.js
jonambas: Yeah good call 👍

packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -1,179 +1,207 @@
...
+      ...style,
+    },
+  };
 
-    const labelMarkup = (
-      <Label id={id} label={`${label}${requiredIndicator}`}>
-        {error && errorInLabel && <Error className={styles.InlineError} wrapper='span' error={error} />}
+  return (
+    <StyledWrapper {...systemProps}>
+      <Label id={id} label={label} labelHidden={labelHidden}>
```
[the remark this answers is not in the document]
jonambas: Yup

packages/matchbox/src/components/TextField/TextField.js
jonambas: Added this one onto the `style` object to avoid pulling in styled-components for one rule

packages/matchbox/src/components/TextField/TextField.js
jonambas: Sure, if we think it wont be too much of a hassle at this point. I think most TextFields use the redux-form wrapper which handles id nicely.

packages/matchbox/src/components/TextField/TextField.js
jonambas: Hmm yea let me revert this

jonambas: Actually I think I have to leave this one in. Otherwise any `margin` system props passed onto the wrapper will be duplicated on the input

jonambas:
@nicklemmon Think I hit everything here – ended up having to write a custom `omit` function

packages/matchbox/src/components/TextField/TextField.js
nicklemmon: 👍 I added it to the story. It shouldn't be a big hassle.

packages/matchbox/src/components/TextField/TextField.js
nicklemmon: Screen Shot 2020-03-13 at 11 36 33 AM

Hmm, now the text is overlapping. Probably because of those padding adjustments you made?

jonambas:
> @jonambas Everything looks good except I think that padding adjustment introduced a bug:
> Screen Shot 2020-03-13 at 11 36 33 AM

Really want to write a test for this but DOM layout calculations are going to be really flakey

PR #351, opened 2020-03-09
FE-913 Move Global Styles to Styled Components

jonambas:
### What Changed
– Global styles are no longer bundled into a styles.css export
– Global styles are generated through ThemeProvider

### How To Test or Verify
- Verify global styles are working in any story

### PR Checklist


packages/matchbox/src/index.js
```diff
@@ -1,2 +1 @@
-import './styles/global.scss';
```
bkemper: Should we delete this file?

bkemper: Or, is this saved, so you could use it with `skipGlobalStyles`?

packages/matchbox/src/components/ThemeProvider/globalStyles.js
```diff
@@ -0,0 +1,141 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+export default `
+  html {
+    position: relative;
+    box-sizing: border-box;
+    font-size: 16px;
```
bkemper: Like we were talking about in the change, should we use `$DEFAULT_FONT_SIZE`?

stories/storyHelpers/StoryContainer.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import '../../packages/matchbox/styles.css';
```
bkemper: 🥇

packages/matchbox/src/components/ThemeProvider/globalStyles.js
jonambas: We should, but the JS tokens don't export a default font size, just the scale. This is one token we should add though.

jonambas: I'll cut a ticket to add a new token

PR #353, opened 2020-03-12
FE-811 Update Toggle Component

logansparlin:

### What Changed

- Restyles Toggle Component
- Deprecates compact prop

### How To Test or Verify
- `npm run start:storybook`
- compare design to https://sparkpost.invisionapp.com/share/FJUVWXX7BXP#/395966938_Form_Inventory

### PR Checklist


packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -1,68 +1,75 @@
...
+      <StyledOutline checked={checked} />
+      <StyledIndicator checked={checked} />
+    </StyledToggle>
+  );
 }
 
+Toggle.displayName = 'Toggle';
+Toggle.propTypes = {
+  checked: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
+  compact: deprecate(PropTypes.bool, 'Compact prop has been removed', 'error'),
```
nicklemmon: Shouldn't this be a warning? I thought that's how we were handling these.

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -1,68 +1,75 @@
...
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
+import { visuallyHidden } from '../../styles/helpers';
+import styled from 'styled-components';
+import { deprecate } from '../../helpers/propTypes';
+import { ScreenReaderOnly } from '../ScreenReaderOnly';
+import { margin } from 'styled-system';
+import { createPropTypes } from '@styled-system/prop-types';
+import { toggle, input, StyledIndicator, StyledOutline } from './styles';
+import { pick } from '@styled-system/props';
```
nicklemmon: Nitpicky, but with the imports here, could we move the NPM imports to the top?

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -1,68 +1,75 @@
...
+function Toggle(props) {
+  const {
+    id,
+    value,
+    checked,
+    disabled,
+    onChange,
+    onFocus,
+    onBlur,
+    label,
```
nicklemmon: This is a new prop for the old version of the component - I guess we shouldn't pass it on in that case? Should I add that to the acceptance criteria for the component wrapper?

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -1,68 +1,75 @@
...
+    required,
```
nicklemmon: Looks like both `label` and `required` are in the same boat as far as props that didn't exist on the old version?

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -1,68 +1,75 @@
...
-          {...rest} />
-        <span className={styles.Outline}></span>
-        {labelMarkup}
-        <span className={styles.Indicator}></span>
-      </label>
-    );
-  }
+  return (
+    <StyledToggle htmlFor={id} disabled={disabled} {...systemProps}>
+      {label && <ScreenReaderOnly>{label}</ScreenReaderOnly>}
```
nicklemmon: Hmm, this won't always work, as sometimes the label is actually visible - _but_ we can save that for when component APIs update

packages/matchbox/src/components/Toggle/styles.js
```diff
@@ -0,0 +1,68 @@
+import { tokens } from '@sparkpost/design-tokens';
+import styled from 'styled-components';
+
+export const toggle = props => `
+  display: inline-block;
+  position: relative;
+  height: 1.25rem;
+  width: 2.25rem;
```
nicklemmon: You probably already thought of this, but would it be possible to leverage design tokens here? May not actually work well visually, so totally fine if not.

packages/matchbox/src/components/Toggle/styles.js
```diff
@@ -0,0 +1,68 @@
...
+
+  opacity: ${props.disabled ? '0.6' : '1'}
+
+  &:hover {
+    cursor: ${props.disabled ? 'not-allowed' : 'pointer'};
```
nicklemmon: If you wanted, you could just move this outside of the `:hover`? I think it would still work and decrease specificity of the rule slightly.

packages/matchbox/src/components/Toggle/styles.js
```diff
@@ -0,0 +1,68 @@
...
+export const indicator = props => `
+  pointer-events: none;
+  position: absolute;
+  top: 1px;
+  left: 1px;
+  bottom: 1px;
+  width: calc(2.25rem / 2);
+
+  background: ${tokens.color_white};
+  transition: ${tokens.motionDuration_fast} ${tokens.motionEase_in_out};
```
nicklemmon: Better to not transition `all` here for perf reasons - it may be repetitive, but listing out each property that actually transitions is the better way to go.

packages/matchbox/src/components/Toggle/styles.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+`;
+
+export const indicator = props => `
+  pointer-events: none;
+  position: absolute;
+  top: 1px;
+  left: 1px;
+  bottom: 1px;
+  width: calc(2.25rem / 2);
```
nicklemmon: This width value could maybe be stored as a variable to make managing this more clear?

packages/matchbox/src/components/Toggle/styles.js
```diff
@@ -0,0 +1,68 @@
...
+    box-shadow: 0 0 0 2px white, 0 0 0 4px ${tokens.color_blue_600};
+  }
+
+  &:checked {
+    & ~ ${StyledIndicator} {
+      transform: translate(${tokens.spacing_400}, 0);
+    }
+
+    & ~ ${StyledOutline} {
+      background: ${tokens.color_green_600};
```
nicklemmon: This is a slightly unusual situation, but unfortunately this isn't meeting color contrast guidelines - `#fff` on `#28bd8b` doesn't pass.

There are a couple of ways to handle this:
1. Darken the green a lot
2. Make the box shadow on the switch a good bit darker, or add an outline to the white switch element that meets the criteria

Then technically, if the darkest part of the box shadow passes against `#28bd8b` then you're good. Had a very similar situation with a very similar component a few months ago.

packages/matchbox/src/components/Toggle/tests/Toggle.test.js
```diff
@@ -1,37 +1,48 @@
...
+    expect(wrapper.find('span').at(1)).toHaveStyleRule('transform', 'translate(1rem,0)');
+  });
 
-  cases('renders toggle states', (opts) => {
-    wrapper.setProps(opts.props);
-    expect(wrapper).toMatchSnapshot();
-  }, testCases);
+  it('renders default disabled styles', () => {
+    wrapper = subject({ disabled: true });
+    expect(wrapper).toHaveStyleRule('opacity', '0.6');
```
nicklemmon: Can we check that the `disabled` attribute renders on the markup?

nicklemmon: Similarly for the `required `component?

packages/matchbox/src/components/Toggle/Toggle.js
nicklemmon: Also, for the Storybook stories, could you pass in some content here as an example? Might be good to also have stories for the `required` version

packages/matchbox/src/components/Toggle/Toggle.js
jonambas: Woops, yep I'll remove this – was meant to be in another PR. Had a discussion with Brian in another PR

packages/matchbox/src/components/Toggle/Toggle.js
jonambas: `label` is new, `required` technically existed before through `...rest`. Just wanted to start being explicit props we know we want.

packages/matchbox/src/components/Toggle/Toggle.js
jonambas: This would be part of `ToggleBlock` implementation though right? I think I intended this prop to only be used when another label isn't, eg. toggles inside table rows.

packages/matchbox/src/components/Toggle/styles.js
jonambas: Yeah, I'm realizing we have a gap in our spacing tokens. They were really only meant to handle margin and padding, not things like size. But either way, I think this is fine as long as we are using rems and referencing a baseline font size token (which we have a ticket for)

packages/matchbox/src/components/Toggle/Toggle.js
nicklemmon: That makes sense to me for sure.

PR #354, opened 2020-03-16
FE-926 Specify Button Type

jonambas:
### What Changed
- Specifies `type="button"` on component buttons in Snackbar, Banner, Expandable, Tag
- Also changes the expandable arrow rotations to match mockups

### How To Test or Verify
- Verify these component buttons have type specified
- Optional, use one of these components in an HTML form and verify they do not submit the form. Try wrapping expandable with `<form onSubmit={() => alert('submit')}>` in a story.
  - Try with Safari, I couldn't reproduce with Chrome

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Banner/styles.js
+++ b/packages/matchbox/src/components/Banner/styles.js
@@ -71,6 +71,9 @@ export function dismissBase() {
   return `
     padding: 0.25rem;
     transition: background ${tokens.motionDuration_fast} ${tokens.motionEase_in_out};
+    &:hover {
+      cursor: pointer;
+    }
   `;
 }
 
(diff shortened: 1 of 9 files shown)


packages/matchbox/src/components/Snackbar/Snackbar.js
```diff
@@ -15,6 +16,7 @@ const StyledBox = styled(Box)`
 `;
 
 const StyledClose = styled(Box)`
+  ${buttonReset}
```
jonambas: Without this, there was a visual bug that prevented the icons from using the correct color

bkemper:
Should the banner close button have `cursor: pointer`?

jonambas:
> Should the banner close button have `cursor: pointer`?

Good catch!

packages/matchbox/src/components/Banner/styles.js
```diff
@@ -71,6 +71,9 @@ export function dismissBase() {
   return `
     padding: 0.25rem;
     transition: background ${tokens.motionDuration_fast} ${tokens.motionEase_in_out};
+    &:hover {
```
bkemper: Why is this style prop only set on hover?

jonambas: No idea, I've always done it because it feels more semantically correct. But either way don't think it matters too much

PR #356, opened 2020-03-19
FE-819 Update Page Component

jonambas:
### What Changed
- Restyles the Page component
- Does not touch Popover, EmptyState, or Actionlist, which are all dependencies of Page
- Removes the default prop `empty: {}`

Screen Shot 2020-03-19 at 3 16 43 PM

### How To Test or Verify
- Visit and verify Page stories

### PR Checklist


packages/matchbox/src/components/Page/Page.js
```diff
@@ -1,149 +1,218 @@
...
+      trigger={
+        <Box position="relative" mr={hasPrimaryAction ? '200' : ' 0'}>
+          <Button
+            outline
+            color="blue"
+            width="2.5rem" // Forces a square
+          >
+            <Box position="absolute">
+              <MoreHoriz />
+              <ScreenReaderOnly>More Options</ScreenReaderOnly>
```
jonambas: Need suggestions on what this label should be

packages/matchbox/src/components/Page/Page.js
```diff
@@ -1,149 +1,218 @@
...
+import { Box } from '../Box';
+import { Button } from '../Button';
 import { EmptyState } from '../EmptyState';
-import { UnstyledLink, linkFrom } from '../UnstyledLink';
+import { Popover } from '../Popover';
+import { UnstyledLink } from '../UnstyledLink';
+import { ScreenReaderOnly } from '../ScreenReaderOnly';
 import { filterByVisible } from '../../helpers/array';
 
-import styles from './Page.module.scss';
```
bkemper: Shouldn't this file be removed?

packages/matchbox/src/components/Page/Page.js
```diff
@@ -1,149 +1,218 @@
...
+const Wrapper = styled('div')`
+  ${margin}
+`;
+
+function Breadcrumb({ content, ...rest }) {
+  return (
+    <UnstyledLink {...rest} fontSize="200" lineHeight="200" fontWeight="medium">
```
bkemper: Looks like these props, font-size and font-weight, get leaked through to the DOM.

Screen Shot 2020-03-19 at 5 09 56 PM

packages/matchbox/src/components/Page/Page.js
bkemper: I think that works.

packages/matchbox/src/components/Page/Page.js
```diff
@@ -1,149 +1,218 @@
...
-
-    /**
-      * Main cta. Most button props will work in here.
-      * e.g. { content: 'button label', onClick: callback() }
-      */
-    primaryAction: PropTypes.shape({
-      content: PropTypes.node.isRequired
-    }),
+  return (
+    <Popover
```
bkemper: Won't this need some aria property to connect it with the button?

jonambas: I think so, but can address it in the popover ticket. I believe all it needs is an `id` and `describedby`, but probably also a focus handler

packages/matchbox/src/components/Page/Page.js
jonambas: Styled components doesn't have a way to omit props – they do have a `shouldForwardProps` utility that should be released soon 🤞

packages/matchbox/src/components/Page/Page.js
jonambas: Yep thanks

PR #359, opened 2020-03-22
FE-919 - Updates ScreenReaderOnly to use styled components

nicklemmon:
[FE-919](https://jira.int.messagesystems.com/browse/FE-919)

### What Changed
- Updates `ScreenReaderOnly` to use styled components instead of CSS modules
- Adds support for `as` prop
- Adds support for `id` prop
- Updates unit tests accordingly
- Re-formats stories file with Prettier and removes descriptive text to match more recent examples

### How To Test or Verify
- Verify that content within the component is in the DOM but not visible

### PR Checklist
- [X] Add your changes to `unreleased.md` in the root directory. If you plan on publishing immediately after merging (such as a hotfix) or aren't making changes to a published package, you can ignore this step.
- [ ] Incorporate feedback


packages/matchbox/src/components/ScreenReaderOnly/ScreenReaderOnly.js
```diff
@@ -1,15 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styles from './ScreenReaderOnly.module.scss';
+import styled from 'styled-components';
+import { visuallyHidden } from '../../styles/helpers';
+
+const StyledScreenReaderOnly = styled('span')`
```
bkemper: How does this work when we used block element inside it?

https://github.com/SparkPost/2web2ui/blob/master/src/components/hibana/DesktopNavigation.js#L29-L31

nicklemmon: You are definitely right - we should expose an `as` prop. I'll go ahead and do that as well so then I can incorporate it as part of the component wrapper ticket as well.

PR #360, opened 2020-03-23
FE-941 Add root font size token

jonambas:
### What Changed
- Added some very basic tests to verify the token build works as expected
- Publishes design tokens
- Uses the new root font size token in matchbox global styles

### How To Test or Verify
- Inspect the page content, verify `html` has `16px` set as a font size

### PR Checklist


packages/design-tokens/package.json
```diff
@@ -15,18 +15,15 @@
   },
   "scripts": {
     "prebuild": "npm run clean",
-    "build": "npm-run-all build:**",
-    "build:gulp": "gulp",
-    "build:scss": "node scripts/scss.js",
-    "build:format": "prettier ./{dist,docs}/*.{js,json,css,scss} --write --no-config",
+    "build": "gulp && node scripts/scss.js && npm run format",
```
nicklemmon: These change makes the root project not work when running `npm install` - not totally clear why. Looks like `gulp` isn't listed as a root dependency?

jonambas: `lerna bootstrap` or prebuild _should_ be taking care of it. Its just a bit weird since we also require a `lerna clean` in pretest to remove duplicate versions of react. Let me see if adding gulp and theo to the workspace help

PR #361, opened 2020-03-24
FE-950 & FE-983 – Prevent tab content from wrapping, Specify tab button type, Fix component wrappers

jonambas:
### What Changed
- Prevents tab text from wrapping for tabs with multiple words
- Specifies button `type` to prevent forms from submitting when clicking tabs
- Fixes instances where `component` or `Component` pass through a link

### How To Test or Verify
- Visit tab stories, verify text does not wrap
- Optionally, you can wrap tabs in a `form` with an onSubmit to check if clicking a tab submits the form


jonambas:
@shrirupa Thanks!

PR #362, opened 2020-03-25
FE-821 Update Pagination Component

logansparlin:

### What Changed
- Restyle the pagination component
- Deprecated selectedColor and flat props
- Add margin system props

### How To Test or Verify
- `npm run start:storybook`
- compare to designs: https://sparkpost.invisionapp.com/share/FJUVWXX7BXP#/screens/393107481_Button_Symbols
### PR Checklist


packages/matchbox/src/components/Pagination/tests/Pagination.test.js
```diff
@@ -1,65 +1,57 @@
...
-
-    beforeEach(() => {
-      fn = jest.fn();
-      wrapper.setProps({ onChange: fn });
-    });
-
     it('page', () => {
       wrapper.setProps({ pages: 11 });
-      expect(fn).toHaveBeenCalledTimes(1);
+      expect(props.onChange).toHaveBeenCalledTimes(1);
```
jonambas: Just to make sure we aren't breaking anything, could we assert on `toHaveBeenCalledWith`?

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -128,55 +80,107 @@ class Pagination extends Component {
...
+
+  const lastButton =
+    !marginsHidden && start + pageRange < pages ? (
+      <span>
+        <Box display="inline" pl={200} pr={200}>
+          <StyledMoreHoriz size={24} />
+        </Box>
+        <Button
+          flat
+          width={index + 1 < 100 ? '32px' : 'auto'}
```
jonambas: Should we use a `rem` value here for 32px?

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -1,119 +1,71 @@
...
-  handleNext = () => {
-    this.handlePageChange(this.state.index + 1);
-  };
-
-  handlePrevious = () => {
-    this.handlePageChange(this.state.index - 1);
-  };
-
-  _createButtons(start, end) {
+  function _createButtons(start, end) {
```
jonambas: This is a really good opportunity to try out `React.useMemo` if you haven't used it yet. I can walk you through it if you'd like, but it is essentially the same as `useEffect` except you return a value instead. Great for functions like.

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -1,119 +1,71 @@
...
-          index === this.state.index && styles.Selected,
-          index === this.state.index && styles[`color-${this.props.selectedColor}`],
-        ),
+        content: `${currentIndex + 1}`,
+        onClick: handlePageChange.bind(this, currentIndex),
+        flat: currentIndex === index ? false : true,
+        size: 'small',
+        color: currentIndex === index ? 'blue' : 'gray',
+        mx: '100',
+        width: currentIndex + 1 < 100 ? '32px' : 'auto',
```
jonambas: Same comment here with `32px`

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -1,119 +1,71 @@
...
       });
     }
 
     return buttonsFrom(buttons);
```
jonambas: Curious if you think we should keep using this function – I don't really have too much of an opinion but some prefer rendering JSX

packages/matchbox/src/components/Pagination/tests/Pagination.test.js
```diff
@@ -1,65 +1,57 @@
...
     });
 
     it('currentPage', () => {
       wrapper.setProps({ currentPage: 4 });
-      expect(fn).toHaveBeenCalledTimes(1);
+      expect(props.onChange).toHaveBeenCalledTimes(1);
```
jonambas: Is it possible to also assert the right button has the right active styles?

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -48,21 +48,23 @@ function Pagination(props) {
...
-        flat: currentIndex === index ? false : true,
-        size: 'small',
-        color: currentIndex === index ? 'blue' : 'gray',
-        mx: '100',
-        width: currentIndex + 1 < 100 ? '32px' : 'auto',
-      });
-    }
-
-    return buttonsFrom(buttons);
+    return React.useMemo(() => {
```
jonambas: So React.useMemo shouldn't be used in a callback like this, i think this is what i was imagining:
```js
const buttonMarkup = React.useMemo(() => {
if (pages <= pageRange) {
     return _createButtons(0, pages);
    } else {
      return _createButtons(start, start + pageRange);
    }
}, [start, pages, pageRange])
```
Could even pull in `_createButtons` if you want

jonambas: Sorry for the confusion, added the initial comment to the wrong piece of code

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -114,7 +120,7 @@ function Pagination(props) {
         </Box>
         <Button
           flat
-          width={index + 1 < 100 ? '32px' : 'auto'}
+          width={index + 1 < 100 ? tokens.spacing_600 : 'auto'}
```
jonambas: I think this needs to look at `pages` right not index?

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -97,7 +99,11 @@ function Pagination(props) {
   const firstButton =
     !marginsHidden && start > 1 ? (
       <span>
-        <Button flat width={index + 1 < 100 ? '32px' : 'auto'} onClick={() => handlePageChange(0)}>
+        <Button
+          flat
+          width={index + 1 < 100 ? tokens.spacing_600 : 'auto'}
```
jonambas: Probably not needed if this is always `1`

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -48,21 +48,23 @@ function Pagination(props) {
...
+          onClick: handlePageChange.bind(this, currentIndex),
+          flat: currentIndex === index ? false : true,
+          size: 'small',
+          color: currentIndex === index ? 'blue' : 'gray',
+          mx: '100',
+          width: currentIndex + 1 < 100 ? tokens.spacing_600 : 'auto',
+        });
+      }
+
+      return buttonsFrom(buttons);
```
jonambas: Not really part of this ticket but noticed buttons are missing `cursor: pointer` 😁

packages/matchbox/src/components/Pagination/Pagination.js
```diff
@@ -128,55 +82,111 @@ class Pagination extends Component {
...
+          width={index + 1 < 100 ? tokens.spacing_600 : 'auto'}
+          onClick={() => handlePageChange(pages - 1)}
+        >
+          {pages}
+        </Button>
+      </span>
+    ) : null;
+
+  return (
+    <Box display="inline-flex" alignItems="center" {...rest}>
```
jonambas: using Box here will allow all system-props through, not just margin. You could just add another div wrapper aound this with those system props

stories/navigation/Pagination.stories.js
```diff
@@ -1,35 +1,44 @@
...
-    <div data-id="pagination-flat">
-      <Pagination
-        flat
-        pages={30}
-        pageRange={3}
-        selectedColor='navy'
-        onChange={action('Page Changed')}
-      />
+export const WithLotsOfPages = withInfo()(() => (
+  <div data-id="pagination-lots-of-pages">
```
jonambas: in `Pagination.spec.js` I think you might be able to replace `<div data-id="pagination-flat">` with `pagination-lots-of-pages` and the test will pass. Looks like the pagination props are exactly the same, just missing `flat`.

You may need to update the iframe url too since the story title changed

PR #363, opened 2020-03-26
FE-942 - Resolve label bug by conditionally rendering label when none is passed

nicklemmon:
[FE-942](https://jira.int.messagesystems.com/browse/FE-924)

### What Changed
- Conditionally renders label component within the `<TextField/>` component

### How To Test or Verify
- In Storybook, remove a label from the basic example and verify in the DOM that it does not render.
- Checkout `master` and verify that the label _is_ rendering (to reproduce the bug)

### PR Checklist
- [ ] Incorporate feedback


packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -139,16 +139,18 @@ function TextField(props) {
...
-        {required && (
-          <Box as="span" pr="200" aria-hidden="true">
-            *
-          </Box>
-        )}
-        {error && errorInLabel && (
-          <Box as={Error} id={errorId} wrapper="span" error={error} fontWeight="400" />
-        )}
-      </Label>
+      {label && (
```
jonambas: Looks like Selects could use this too. Would it make more sends to add `if (!label) return null` in the `Label` component? Up to you here

nicklemmon: that's a good idea!

nicklemmon: for the record, it looks like Select was already handling this.

PR #365, opened 2020-03-27
FE-981 Fix Tooltip positioning

jonambas:
### What Changed
- Duplicates tooltip content into a screen reader only element next to triggers
- Adds a new `as` prop to Tooltips, defaults to `span` – Realized the old implementation forced an inline-block `span`, so this should not be a breaking change
- Tooltips now visible if trigger is focused

### How To Test or Verify
- Verify Tooltip stories and that tooltips are positioned correctly

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Tooltip/Tooltip.js
+++ b/packages/matchbox/src/components/Tooltip/Tooltip.js
@@ -105,6 +105,7 @@ function Tooltip(props) {
     return (
       <Box
         as={as}
+        // Inline block is required to measure and set height
         display={as === 'span' ? 'inline-block' : null}
         onFocus={handleShow}
         onBlur={handleHide}
@@ -121,6 +122,7 @@ function Tooltip(props) {
     <TooltipOverlay
       id={props.id}
       hideTooltip={handleHide}
+      portalId={props.portalId}
       renderA11yContent={renderA11yContent}
       renderTooltip={renderTooltip}
       renderActivator={renderActivator}
@@ -132,7 +134,7 @@ function Tooltip(props) {
 Tooltip.displayName = 'Tooltip';
 Tooltip.propTypes = {
   as: PropTypes.string,
-  id: PropTypes.string,
+  id: PropTypes.string.isRequired,
   content: PropTypes.node,
   /**
    * Disables hover events
--- a/packages/matchbox/src/components/Tooltip/Tooltip.js
+++ b/packages/matchbox/src/components/Tooltip/Tooltip.js
@@ -32,6 +32,7 @@ function Tooltip(props) {
 
   function renderTooltip({ preferredDirection }) {
     const {
+      as,
       content,
       dark, // TODO deprecate in favor of system props
       top,
@@ -89,9 +90,10 @@ function Tooltip(props) {
   }
 
   function renderActivator({ activatorRef }) {
+    console.log(props.as);
     return (
       <Box
-        display="inline-block"
+        display={props.as === 'span' ? 'inline-block' : null}
         onMouseOver={handleMouseOver}
         onMouseOut={handleMouseOut}
         ref={activatorRef}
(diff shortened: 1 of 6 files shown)


packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -89,11 +100,16 @@ function Tooltip(props) {
...
   function renderActivator({ activatorRef }) {
+    const { as = 'span' } = props;
+
     return (
       <Box
-        display="inline-block"
-        onMouseOver={handleMouseOver}
-        onMouseOut={handleMouseOut}
+        as={as}
+        display={as === 'span' ? 'inline-block' : null}
```
jonambas: Previous implementation forced inline-block and old tooltips never worked correctly on block level elements, this shouldn't be a breaking change

packages/matchbox/src/components/Tooltip/TooltipOverlay.js
```diff
@@ -49,24 +57,29 @@ function TooltipOverlay(props) {
...
+          position="absolute"
+          top="0"
+          left="0"
+          style={{
+            pointerEvents: 'none',
+          }}
+        >
+          <Box
+            position="absolute"
+            top={`${position.top}px`}
```
jonambas: Appending `px` here to avoid matching a token key

bkemper:
> Tooltips now visible if trigger is focused

Very awesome!

Screen Shot 2020-03-31 at 1 28 19 PM

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -89,11 +100,16 @@ function Tooltip(props) {
   }
 
   function renderActivator({ activatorRef }) {
+    const { as = 'span' } = props;
```
bkemper, line 103: I wanted to ask this earlier, but why not use defaultProps?

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -142,7 +160,10 @@ Tooltip.propTypes = {
     right: PropTypes.bool,
   }),
 
-  portalId: deprecate(PropTypes.string, 'Portals are no longer used', 'error'),
+  /**
+   * Element ID for the portal that will house tooltips. Appends to body if not provided.
+   */
+  portalId: PropTypes.string,
```
bkemper, line 166: I don't see this being used.

packages/matchbox/src/components/Tooltip/Tooltip.js
bkemper: This comment should probably be in code because "as" is not restricted to only block elements and I could set "as" to "b" and it won't behave the same.

packages/matchbox/src/components/Tooltip/Tooltip.js
```diff
@@ -20,14 +21,24 @@ const StyledContent = styled('div')`
...
-  function handleMouseOut() {
-    setHover(false);
+  function handleHide() {
+    setshow(false);
+  }
+
+  function renderA11yContent() {
+    return (
+      <ScreenReaderOnly>
+        <span role="tooltip" aria-hidden={!show} id={props.id}>
```
bkemper, line 37: Should "aria-describedby" be set in `renderActivator`?

```
<!-- see, https://www.accede-web.com/en/guidelines/rich-interface-components/tooltips-aria/ -->
<a href="#" aria-describedby="extrainfo">[Link text]</a>
<div role="tooltip" id="extrainfo">[Content of the tooltip]</div>
```

bkemper:
Can we chat about this?

jonambas:
- Make `id` required
- Add comment for inline block
- Pass through `portalId`
- Update stories with describedBy

bkemper:
push build failed because of Slider tests?

Screen Shot 2020-04-01 at 10 17 40 PM

jonambas:
@bkemper its a flakey test – I usually just hit rebuild but I lost that option. Commented it out in another PR

bkemper:
You can also use it.skip.

PR #367, opened 2020-03-31
FE-942 - <Label/> component does not render when no "label" prop is provided

nicklemmon:
[FE-942](https://jira.int.messagesystems.com/browse/FE-942)

### What Changed
- `<Label/>` renders as `null` when no `"label"` is passed in

### How To Test or Verify
- Verify no `<Label/>` renders for the `<TextField/>` or `<Select/>` when no `"label"` is provided.

### PR Checklist
- [ ] Incorporate feedback

### DIFF
--- a/packages/matchbox/src/components/Select/Select.js
+++ b/packages/matchbox/src/components/Select/Select.js
@@ -123,7 +123,7 @@ function Select(props) {
 
   return (
     <StyledWrapper {...systemProps}>
-      {label && labelMarkup}
+      {labelMarkup}
       <Box position="relative">
         <StyledSelect
           id={id}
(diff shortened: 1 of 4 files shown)


nicklemmon:
@jonambas Had to close the old one - I goofed and rebased it off `master` - but here we go. Ready for another pass.

packages/matchbox/src/components/Select/Select.js
```diff
@@ -123,7 +123,7 @@ function Select(props) {
 
   return (
     <StyledWrapper {...systemProps}>
-      {label && labelMarkup}
```
jonambas: Oh heh 😓

PR #369, opened 2020-03-31
FE-826 Update Popover Component

jonambas:
### What Changed
- Restyles `Popover`
- Adds new `id` prop for `Popover`
- Adds new `as` prop for `Popover`, functions the same way as `wrapper`
- Removes the `portalId` prop from `Popover`
- Removes the `fixed` prop from `Popover`, since popover content no longer uses portals
- Adds new exported hook `useWindowEvent`

Screen Shot 2020-04-01 at 7 58 40 PM

### How To Test or Verify
- Verify popover stories
- Take a look at cypress tests and verify all window events are tested
- Verify other components or stories that use Popover – Page and Actionlist

### PR Checklist


packages/matchbox/src/components/Popover/Popover.js
```diff
@@ -1,159 +1,152 @@
...
+  /**
+   * Callback function that is called when clicking outside the popover, or hitting escape.
+   */
+  onClose: PropTypes.func,
+  /**
+   * Popover Content
+   */
+  children: PropTypes.node,
+  as: PropTypes.oneOf(['div', 'span']),
+  wrapper: PropTypes.oneOf(['div', 'span']),
```
jonambas, line 147: `as` and `wrapper` here are the same thing, but figured I'd support `as` since that seems to be the convention these days – Tooltip is also using the same prop name.

Not sure about restricting this prop to div and span only

packages/matchbox/src/components/Popover/tests/Popover.test.js
```diff
@@ -1,175 +1,128 @@
 import React from 'react';
```
jonambas, line 1: Removed window event tests from this file – They are now a part of cypress tests, but are not reflected in coveralls / test coverage

cypress/integration/Slider.spec.js
```diff
@@ -13,20 +13,20 @@ describe('The Slider component', () => {
 
     // Test relies on timing for layout side effects
     // Remove if flakey
-    it('should update the sliders value when clicking on the track', () => {
-      cy.wait(200); // Gives slider some time for calculations
+    // it('should update the sliders value when clicking on the track', () => {
```
jonambas, line 16: This test fails half too often – commenting it out

packages/matchbox/src/components/Popover/Popover.js
```diff
@@ -1,159 +1,153 @@
...
   }
 
-  handleTrigger = () => {
-    if (this.state.open === false) {
-      this.uncontrolledToggle();
+  // Toggles uncontrolled popovers
+  function handleTrigger() {
+    if (open !== null) {
+      handleUncontrolledToggle();
     }
```
nicklemmon, line 68: A few things about the trigger:
1. It looks like `aria-expanded` is missing from the example on the button element...I know this won't be supplied by the component (though I think ideally it would - maybe when we update component APIs)
1. Clicking on the button repeatedly appears to cause a bug? It sort of opens and closes but appears to get stuck in a certain state, especially if you click rapidly.

cypress/integration/Popover.spec.js
```diff
@@ -36,3 +38,43 @@ describe('The Popover component', () => {
     });
   });
 });
+
+describe('Uncontrolled Popover with Actionlist', () => {
+  beforeEach(() => {
+    cy.visit('/iframe.html?path=/story/overlays-popover--with-an-action-list');
+    cy.wait(500); // The element that handles click events requires time to calculate dimensions
```
bkemper, line 45: :(

stories/layout/Page.stories.js
```diff
@@ -28,14 +28,17 @@ const secondaryActions = [
       </span>
     ),
     onClick: action('Save Clicked'),
+    as: 'button',
```
bkemper, line 31: Looks like this will need to be added to the app as well?

packages/matchbox/src/components/Popover/styles.js
```diff
@@ -0,0 +1,46 @@
...
+  margin-bottom: ${props.isTop ? tokens.spacing_100 : tokens.spacing_0};
+
+  left: ${props.isLeft ? 'auto' : '0'};
+  right: ${!props.isLeft ? 'auto' : '0'};
+  top: ${props.isTop ? 'auto' : '100%'};
+  bottom: ${!props.isTop ? 'auto' : '100%'};  
+`;
+
+export const transition = props => {
+  switch (props.state) {
```
bkemper, line 19: When the two forces combine.

packages/matchbox/src/components/Popover/Popover.js
```diff
@@ -1,159 +1,153 @@
...
+    setOpen(!open);
   }
 
-  handleOutsideClick = (e) => {
-    const { open: controlledOpen, onClose } = this.props;
-    const isOutside = this.popover && !this.popover.contains(e.target) && this.activator && !this.activator.contains(e.target);
+  // Toggles uncontrolled popovers, and calls `onClose` for controlled popovers
+  function handleOutsideClick(e) {
+    const isOutside =
+      !popoverRef.current.contains(e.target) && !activatorRef.current.contains(e.target);
```
bkemper, line 40: Since the listener is on the window, these unfortunately need the guards to check if the refs exist.
Screen Shot 2020-04-06 at 1 22 21 PM

packages/matchbox/src/components/Popover/styles.js
```diff
@@ -0,0 +1,46 @@
...
+    case 'entering':
+    case 'entered':
+      return `
+        pointer-events: auto;
+        visibility: visible;
+        opacity: 1;
+        transform: translateY(0);
+        transition: opacity ${tokens.motionDuration_medium} ${tokens.motionEase_out},
```
bkemper, line 27: I think the transition state falls out of sync with the component state which causes the popover to get stuck if you click the activator a couple times.

Screen Shot 2020-04-06 at 1 37 23 PM

packages/matchbox/src/components/Popover/Popover.js
```diff
@@ -1,159 +1,153 @@
...
-    children: PropTypes.node,
-    /**
-     * Element ID for the portal that will house Popover. Appends to body if not provided.
-     */
-    portalId: PropTypes.string
-  };
-
-  static defaultProps = {
-    right: true,
-    bottom: true
```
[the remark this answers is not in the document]
jonambas: Yeah, the directional boolean props I think need to be reworked. eg `bottom={false}` doesn't position the popover on top, and the component only really checks for `left` and `top`. So a snapshot with these default props would be inaccurate.

I think eventually we'll move towards `position="bottomRight|topLeft|etc"`

stories/layout/Page.stories.js
jonambas: yep, it defaults to `link` which is the same as the old implementation

packages/matchbox/src/components/Popover/styles.js
jonambas: I don't think I can reproduce this, but I can try to disable pointer events until content has been fully rendered?

jonambas: I think this might be a bug with storybook actions – can't tell if you have them enabled. Sometimes they freeze the browser

jonambas:
@bkemper Think this ones ready for another pass

packages/matchbox/src/components/Popover/Popover.js
bkemper: I think we will find out.

packages/matchbox/src/components/Popover/Popover.js
bkemper: 1. This was addressed.
2. I saw the same thing.  Jon thinks it is due to Storybook.  Punting for now.

stories/layout/Page.stories.js
```diff
@@ -28,14 +28,17 @@ const secondaryActions = [
       </span>
     ),
     onClick: action('Save Clicked'),
+    is: 'button',
```
bkemper, line 31: Ahh, I see — ActionList.

packages/matchbox/src/components/Page/Page.js
```diff
@@ -47,35 +47,41 @@ function Subtitle({ subtitle }) {
...
+      <Popover
+        bottom
+        id="page-secondary-actions"
+        left
+        onClose={() => setIsOpen(false)}
+        open={isOpen}
+        trigger={
           <Button
-            outline
+            aria-describedby="page-secondary-actions"
```
nicklemmon, line 68: @jonambas Just saw this, but looks like it's already been merged - technically `aria-describedby` isn't really appropriate in this scenario. `aria-controls` would probably make more sense - the `<Expandable/>` component would be a good reference for what that should look like.

jonambas: Eek, thanks that makes sense. I'll cut a follow up ticket

PR #371, opened 2020-04-01
FE-825 - Update Modal Component

nicklemmon:
[FE-825](https://jira.int.messagesystems.com/browse/FE-825)

### What Changed
- Refactors Modal component to use functional instead of class components
- Removes CSS modules in favor of Styled Components
- Removes distinct `<Content/>` component in favor of a private, `<ModalContent/>` component
- Adds supporting unit tests
- Adds mock for `react-transition-group` for use by Jest
- Adds supporting end-to-end tests
- Updates Storybook examples to match mocks, refactors Storybook `<ModalDemo/>` component from a class to a function

### How To Test or Verify
- Run `npm run start:storybook` and verify the modal renders according to the mocks
- Verify the animation works as expected
- Verify unit/e2e tests adequately test the component
- Verify a11y attributes survived the update in their appropriate locations
- Verify the modal is controllable via a keyboard

### PR Checklist
- [ ] Incorporate feedback


cypress/integration/Modal.spec.js
```diff
@@ -2,18 +2,36 @@
 
 describe('The Modal component', () => {
   beforeEach(() => {
```
jonambas, line 4: Should we add a test here to check `focused` is within the modal? Also wondering if it makes sense to render the close button after modal `children` – it seems its will be the first element focused

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -1,88 +1,144 @@
...
 
-  handleKeyDown = (e) => {
-    const { onClose, open } = this.props;
+  return (
+    <Portal containerId={containerId}>
+      <StyledBase
+        {...rest}
+        className={className}
+        onClose={onClose}
+        ref={el => (container = el)}
```
jonambas, line 64: I think you could just do `ref={container}` here with `useRef`

packages/matchbox/src/components/Modal/styles.js
```diff
@@ -0,0 +1,84 @@
...
+  transition: transform ${tokens.motionDuration_medium} ${tokens.motionEase_out},
+              visibility ${tokens.motionDuration_medium} ${tokens.motionEase_out},
+              opacity ${tokens.motionDuration_medium} ${tokens.motionEase_out};
+`;
+
+export const contentAnimation = props => {
+  const commonExitStyles = `
+    transform: translateY(${tokens.spacing_200});
+    opacity: 0;
+    transition: transform ${tokens.motionDuration_fast} ${tokens.motionEase_in},
```
jonambas, line 49: Nice, didn't know you could chain transitions like this

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -1,88 +1,144 @@
...
+  maxWidth: PropTypes.string,
+  /**
+   * Controls the target container ID for the rendering React portal
+   */
+  containerId: PropTypes.string,
+};
+
+Modal.defaultProps = {
+  maxWidth: '800px',
+  containerId: 'modal-portal',
```
jonambas, line 141: Not sure this should have a default, the Portal component defaults to document.body without an id

jonambas: Also Tooltip uses the prop `portalId` if you wanted to align the prop name

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -1,88 +1,144 @@
...
-        <Grid center='xs' middle='xs' className={styles.Grid}>
-          <Grid.Column xs={11} md={9} xl={7}>
-            <Content contentRef={(node) => this.content = node} open={open}>
-              <WindowEvent event='keydown' handler={this.handleKeyDown} />
-              <WindowEvent event='click' handler={this.handleOutsideClick} />
               {showCloseButton && (
-                <Button className={styles.CloseButton} flat onClick={onClose} data-id='modal-close'>
-                  Close <Close />
-                </Button>
+                <StyledCloseButton flat onClick={onClose} data-id="modal-close">
```
jonambas, line 76: :/ Need to add `cursor: pointer` to the Button component – have a ticket for that

packages/matchbox/src/components/Modal/Modal.js
jonambas: Ohhhhh, so because you used `let` with `useRef`, callback refs work and you don't have to use `ref.current` to access it. Wish this was documented better, but it seems doing `const ref = useRef()` changes the behavior of the hook.

cypress/integration/Modal.spec.js
jonambas: Nevermind, didn't see the test below that tabs through

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -1,88 +1,144 @@
...
+
+  return (
+    <StyledFocusLock disabled={!open} maxWidth={maxWidth}>
+      <Transition
+        mountOnEnter
+        unmountOnExit
+        in={open}
+        timeout={{
+          enter: 300,
+          exit: 150,
```
jonambas, line 104: Should this reference the fast token? I know the tokens aren't formatted in milliseconds – up to you, maybe add a comment here

packages/matchbox/src/components/Modal/Modal.js
nicklemmon: Yeeeah I saw that ticket so decided not to add it here.

packages/matchbox/src/components/Modal/Modal.js
nicklemmon: Actually wrote a helper for the app to use that does this. I'll just bring it in to Matchbox too.

packages/matchbox/src/components/Modal/Modal.js
nicklemmon: I was actually mimicking your original implementation in this component. Works well.

jonambas:
Screen Shot 2020-04-02 at 2 25 42 PM

This might be an issue, not sure how long our modal titles are – also don't think this is fixable at the moment

nicklemmon:
> Screen Shot 2020-04-02 at 2 25 42 PM
> 
> This might be an issue, not sure how long our modal titles are – also don't think this is fixable at the moment

Not sure I'm following what the issue is...?

PR #372, opened 2020-04-02
FE-948: Restyle Grid and Grid.Column Components

bkemper:
### What Changed
- Converted Grid and Grid.Column to use styled-components

### How To Test or Verify
- Load Storybook
- Tweak stories
- Confirm Grid and Grid.Columns behavior has not changed

### PR Checklist


packages/matchbox/src/components/Grid/styles.js
```diff
@@ -0,0 +1,25 @@
+import { tokens } from '@sparkpost/design-tokens';
+
+const gutterWidth = tokens.spacing_400;
```
bkemper, line 3: This was the closest token to the original gutter width, `rem(18)`.

Screen Shot 2020-04-02 at 2 50 35 PM

packages/matchbox/src/components/Grid/Column.js
```diff
@@ -1,49 +1,28 @@
...
-      smOffset && styles[`sm-offset-${smOffset}`],
-      mdOffset && styles[`md-offset-${mdOffset}`],
-      lgOffset && styles[`lg-offset-${lgOffset}`],
-      xlOffset && styles[`xl-offset-${xlOffset}`]
-    );
-
-    return <div className={colClasses} {...rest}>{children}</div>;
-  }
+function Column({ children, ...props }) {
+  return <StyledGridColumn {...props}>{children}</StyledGridColumn>;
```
bkemper, line 11: Should we be more restrictive?

packages/matchbox/src/components/Grid/styles.js
```diff
@@ -0,0 +1,103 @@
...
+export const breakpoints = ['xs', 'sm', 'md', 'lg', 'xl'];
+const alignmentStyleMap = {
+  start: {
```
bkemper, line 6: I used objects to represent the style because it was easier to compose together.

packages/matchbox/src/components/Grid/styles.js
```diff
@@ -0,0 +1,103 @@
...
+    display: 'flex',
+    flex: '0 1 auto',
+    flexDirection: 'row',
+    flexWrap: 'wrap',
+    marginLeft: `-${gutterWidth}`,
+    ...breakpointStyle,
+  };
+};
+
+const calculateGridColumnWidth = index => (100 / 12) * index;
```
bkemper, line 70: I don't think there is any big difference in doing this calculation in js vs scss since we are using integers.

packages/matchbox/src/components/Grid/styles.js
```diff
@@ -0,0 +1,103 @@
...
+    const columnStyle = {
+      ...(breakpointIndex
+        ? {
+            flex: `0 0 ${calculateGridColumnWidth(breakpointIndex)}%`,
+            maxWidth: `${calculateGridColumnWidth(breakpointIndex)}%`,
+            paddingLeft: gutterWidth,
+            textAlign: 'left',
+          }
+        : {}),
+      ...(breakpointOffsetIndex
```
bkemper, line 85: Messed around with a couple different ways to make this easier to read and this is what I settled on.  Please provide feedback.

bkemper:
The props for these components allow for users to override each other.  To maintain the current behavior, I matched my loops in styles to the stylesheets — last winning.

packages/matchbox/src/components/Grid/Column.js
jonambas: I see a few instances in the app with `...spread`s on this component – Should probably leave for now

packages/matchbox/src/components/Grid/Grid.js
```diff
@@ -1,51 +1,31 @@
...
-      bottom && styles[`bottom-${bottom}`],
-      around && styles[`around-${around}`],
-      between && styles[`between-${between}`],
-      className
-    );
-
-    return <div className={gridClasses}>{children}</div>;
-  }
+function Grid({ children, ...props }) {
+  return <StyledGrid {...props}>{children}</StyledGrid>;
```
jonambas, line 13: Up to you if you want to restrict this more – doesn't look like the old component spread props

packages/matchbox/src/components/Grid/styles.js
```diff
@@ -0,0 +1,103 @@
...
+  },
+  around: {
+    justifyContent: 'space-around',
+  },
+  between: {
+    justifyContent: 'space-between',
+  },
+};
+
+const wrapStyleWithMediaQuery = (breakpoint, style) => {
```
jonambas, line 35: This is a bit unfortunate :( If we could change the API styled-system has helpers to handle all of this pretty easily

packages/matchbox/src/components/Grid/Column.js
bkemper: I think you are right, but I confirmed the other props that are being set are `key`, `className`, and `style`.

packages/matchbox/src/components/Grid/Grid.js
```diff
@@ -9,8 +10,9 @@ const StyledGrid = styled('div')`
   ${gridStyle}
 `;
 
-function Grid({ children, ...props }) {
-  return <StyledGrid {...props}>{children}</StyledGrid>;
+function Grid(props) {
+  const { children, ...rest } = pick(props, Grid.propTypes);
```
jonambas, line 14: nice

stories/utility/Grid.stories.js
```diff
@@ -1,100 +1,162 @@
...
-  .add('Offset', withInfo({ propTablesExclude: [DemoBlock] })(() => (
-    <Grid>
-      <Grid.Column xsOffset={10} xs={2}><DemoBlock /></Grid.Column>
-      <Grid.Column xsOffset={8} xs={4}><DemoBlock /></Grid.Column>
-      <Grid.Column xsOffset={6} xs={6}><DemoBlock /></Grid.Column>
-      <Grid.Column xsOffset={4} xs={8}><DemoBlock /></Grid.Column>
-      <Grid.Column xsOffset={2} xs={10}><DemoBlock /></Grid.Column>
-    </Grid>
-  )))
+  .addDecorator(getStory => <StoryContainer bg="white">{getStory()}</StoryContainer>)
```
jonambas, line 12: Sorry, I didnt explain correctly – `storiesOf` is deprecated, I believe. The newer storybook usage uses JS exports to define stories. Take a look at `Button.stories.js` for example

PR #375, opened 2020-04-05
FE-982 - Properly pass height to styled component

nicklemmon:
[FE-982](https://jira.int.messagesystems.com/browse/FE-982)

### What Changed
- Properly pass height prop to styled component that was set statically

### How To Test or Verify
- Run Storybook locally and go here: `/iframe.html?id=form-textfield--multiline`
- Verify that on the `rc/4.0.0` and go to the above URL, verify the bug. Checkout `FE-982` and verify the bug is fixed.

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/TextField/TextField.js
+++ b/packages/matchbox/src/components/TextField/TextField.js
@@ -30,7 +30,7 @@ const FieldBox = props => {
       borderRadius="100"
       bg={props.disabled ? 'gray.200' : 'white'}
       lineHeight="2.5rem"
-      height="2.5rem"
+      height={props.height}
       color="gray.900"
       required={props.required}
     />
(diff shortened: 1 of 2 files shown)


packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -30,7 +30,7 @@ const FieldBox = props => {
       borderRadius="100"
       bg={props.disabled ? 'gray.200' : 'white'}
       lineHeight="2.5rem"
-      height="2.5rem"
+      height={props.height}
```
jonambas, line 33: I don't think we previously supported height, I think you might be able to just remove this line. 

_or_, if you wanted to support system props, you'll need to modify this one line
```js
import { margin, height } from 'styled-system';

const componentProps = omit(rest, margin.propNames, height.propNames);
```

jonambas: Just saw this below – nevermind, I see where props.height comes from
```
const inputProps = {
    ...componentProps,
    as: multiline ? 'textarea' : 'input',
    height: multiline ? 'auto' : '2.5rem',
```

PR #380, opened 2020-04-10
FE-987 Set checked prop in toggle

logansparlin:

### What Changed
- Correctly sets checked prop in toggle
- Adds defaultChecked prop

### How To Test or Verify
- `npm run start:storybook`
- verify tests are passing
- verify toggle is being checked by checked prop

### PR Checklist


packages/matchbox/src/components/Toggle/tests/Toggle.test.js
```diff
@@ -18,7 +18,7 @@ describe('Toggle', () => {
   });
 
   it('renders default checked styles', () => {
-    wrapper = subject({ checked: true });
+    wrapper = subject({ checked: true, onChange: jest.fn() });
```
bkemper, line 21: Is this needed?

packages/matchbox/src/components/Toggle/Toggle.js
```diff
@@ -42,7 +43,8 @@ function Toggle(props) {
       <StyledInput
         id={id}
         value={value}
-        defaultChecked={checked}
+        defaultChecked={defaultChecked}
```
bkemper, line 46: Nothing to fix now, but when fixing up the stories, it would be nice to have an example of how this works and it looks like the decorator is missing the StoryContainer.

packages/matchbox/src/components/Toggle/tests/Toggle.test.js
logansparlin: Yea, onChange is required when a controlled value is set so it throws an error if it's not there

PR #381, opened 2020-04-13
FE-997 Add modal container vertical padding, Fix white background ove…

jonambas:
### What Changed
- Adds vertical padding to modal containers, so content never touchers the viewport edges
- (extra) Adds border radius to panels – background color was previously overflowing past borders

Screen Shot 2020-04-13 at 9 58 32 AM

### How To Test or Verify
- Visit modal stories and resize viewport vertically
- Verify modal content does not touch the top of the browser

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Modal/styles.js
+++ b/packages/matchbox/src/components/Modal/styles.js
@@ -10,6 +10,7 @@ export const base = props => {
     left: 0;
     width: 100vw;
     height: 100vh;
+    padding: ${tokens.spacing_700} 0;
     min-height: 100%;
     max-height: 100vh;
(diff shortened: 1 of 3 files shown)


packages/matchbox/src/components/Modal/styles.js
```diff
@@ -10,6 +10,7 @@ export const base = props => {
     left: 0;
     width: 100vw;
     height: 100vh;
+    padding: ${tokens.spacing_700} 0;
```
logansparlin, line 13: Do we want this to be configurable?

jonambas: Probably not – since this is more of a layout component. Similar to page, its probably better to be restrictive

PR #383, opened 2020-04-13
FE-1007 Manually pick Grid props

jonambas:
### What Changed
- Manually picks grid props – the propeTypes reference was causing a bug seen here: https://nostalgic-stonebraker-a9b0c1.netlify.com/?path=/story/utility-grid--responsive
- `babel-preset-react-app` applies `babel-plugin-transform-react-remove-prop-types`
- Unsure if App production build is affected

### How To Test or Verify
- Verify extra props are omitted

### PR Checklist


packages/matchbox/src/components/Grid/Grid.js
```diff
@@ -1,18 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { pick } from '../../helpers/propTypes';
```
bkemper, line 4: I think this is the only place using this function.  It should probably be deleted.

PR #385, opened 2020-04-16
FE-1018 Support CSS injection DOM node targeting

jonambas:
### What Changed
- Adds a new `target` prop to `ThemeProvider`, see https://styled-components.com/docs/api#stylesheetmanager

### How To Test or Verify
- Testing this in storybook is difficult due to our setup, you will need to remove all but one story and make sure only a single story uses `addDecorator`. (We have a follow up ticket to clean up stories)
- In a single `addDecorator` add:

```js
addDecorator(storyFn => (
   <ThemeProvider target={document.getElementById('styles-target')}>
      {storyFn()}
   </ThemeProvider>
));
```

- Verify the story works

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
+++ b/packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
@@ -41,7 +41,7 @@ Manager.displayName = 'MatchboxStyleSheetManager';
  */
 function Theme(props) {
   return (
-    <Manager>
+    <Manager target={props.target}>
       <ThemeProvider theme={{ ...theme, ...props.theme }}>
         {/*
           FE-913:`skipGlobalStyles` is to exclude global styles from unit tests.
@@ -60,6 +60,7 @@ function Theme(props) {
 Theme.displayName = 'MatchboxThemeProvider';
 Theme.propTypes = {
   skipGlobalStyles: PropTypes.bool,
+  target: PropTypes.object,
 };
 
 export default Theme;
--- a/packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
+++ b/packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
@@ -11,6 +11,27 @@ const GlobalStyle = createGlobalStyle`
 `;
 
 /**
+ * See https://styled-components.com/docs/api#stylesheetmanager
+ * Target provides an alternate DOM node to inject styles info
+ *
+ * Example:
+ * In your app's <head>:
+ * <style id="styled-components-target"></style>
+ *
+ * When rendering ThemeProvider:
+ * <ThemeProvider target={document.getElementById('styled-components-target')} />
+ * */
+function Manager({ target, children }) {
+  if (!target) {
+    return children;
+  }
+
+  return <StyleSheetManager target={target}>{children}</StyleSheetManager>;
+}
+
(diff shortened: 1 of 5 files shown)


packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -18,22 +39,25 @@ const GlobalStyle = createGlobalStyle`
...
-      {props.children}
-    </ThemeProvider>
+    <Manager>
+      <ThemeProvider theme={{ ...theme, ...props.theme }}>
+        {/*
+          FE-913:`skipGlobalStyles` is to exclude global styles from unit tests.
+          The custom assertion `toHaveAttributeValue` fails with this turned on.
+          We should either move away from using this assertion (or even enzyme),
+          or try to resolve the error
+        */}
```
nicklemmon, line 51: I think the time to move to React Testing Library is at hand!

nicklemmon: (I know this comment wasn't introduced with this ticket, but if you're ever interested in swapping let me know - I can walk you through it)

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -18,22 +39,25 @@ const GlobalStyle = createGlobalStyle`
...
-      {/* 
-        FE-913:`skipGlobalStyles` is to exclude global styles from unit tests.
-        The custom assertion `toHaveAttributeValue` fails with this turned on.
-        We should either move away from using this assertion (or even enzyme),
-        or try to resolve the error
-      */}
-      {!props.skipGlobalStyles && <GlobalStyle />}
-      {props.children}
-    </ThemeProvider>
+    <Manager>
```
nicklemmon, line 44: Doesn't look like it's working...the `target` prop isn't being passed down here? Should have `<Manager target={props.target}/>`, no?

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
```diff
@@ -18,22 +39,25 @@ const GlobalStyle = createGlobalStyle`
...
+        {!props.skipGlobalStyles && <GlobalStyle />}
+
+        {props.children}
+      </ThemeProvider>
+    </Manager>
   );
 }
 
+Theme.displayName = 'MatchboxThemeProvider';
 Theme.propTypes = {
```
nicklemmon, line 61: Should probably include:
```
target: PropTypes.element
```

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
jonambas: Heh, yea I kind of regret not making the switch before we started all the component rework

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
jonambas: Woops, thanks for catching

packages/matchbox/src/components/ThemeProvider/ThemeProvider.js
jonambas: `element` includes only react elements :/
Added `object` here, although it isnt very helpful

PR #390, opened 2020-04-24
FE-996 Tab Updates

jonambas:
### What Changed
-  `Tabs` are now responsive for both fitted and non-fitted variations
-  `Tab` now uses a `pointer` cursor on hover
-  `Tabs` now accept the `borderBottom` system prop
-  Follow up ticket for keyboard navigation: FE-1067

### How To Test or Verify
- Visit Tab stories, verify responsive behavior

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Tabs/Tabs.js
+++ b/packages/matchbox/src/components/Tabs/Tabs.js
@@ -156,7 +156,7 @@ function Tabs(props) {
                   aria-expanded={popoverOpen}
                   data-id="tab-options-button"
                   flat
-                  onClick={() => setPopoverOpen(true)}
+                  onClick={() => setPopoverOpen(!popoverOpen)}
                 >
                   <MoreHoriz size={20} />
                   <ScreenReaderOnly>More Options</ScreenReaderOnly>
--- a/packages/matchbox/src/components/Tabs/Tabs.js
+++ b/packages/matchbox/src/components/Tabs/Tabs.js
@@ -2,6 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { MoreHoriz } from '@sparkpost/matchbox-icons';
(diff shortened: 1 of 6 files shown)


packages/matchbox/src/components/Tabs/Tabs.js
```diff
@@ -66,14 +88,87 @@ function Tabs(props) {
...
+              bottom
+              open={popoverOpen}
+              onClose={() => setPopoverOpen(false)}
+              trigger={
+                <Button
+                  aria-controls="tab-options"
+                  aria-expanded={popoverOpen}
+                  data-id="tab-options-button"
+                  flat
+                  onClick={() => setPopoverOpen(true)}
```
logansparlin, line 159: Not necessary, but might improve UX to toggle open and closed when clicking this button. As it is right now I can click to open but have to click off of the button to close.

stories/navigation/Tabs.stories.js
```diff
@@ -14,8 +14,8 @@ const tabs = [
     onClick: action('Details Clicked'),
   },
   {
-    content: 'Keys',
-    onClick: action('Keys Clicked'),
+    content: 'More Details',
+    onClick: action('More Details Clicked'),
```
logansparlin, line 18: Would it be possible here to mock out an example of how it functions when a new tab is selected?

packages/matchbox/src/components/Tabs/Tabs.js
jonambas: Hmm yea good catch

stories/navigation/Tabs.stories.js
jonambas: There is a very short example in the stories. But isn't that great. We should probably add a `useTabs` hook later to handle this state
```js
const Example = () => {
  const [i, seti] = React.useState(0);
  return <Tabs selected={i} onSelect={ind => seti(ind)} tabs={tabs} />;
};
```

PR #410, opened 2020-05-11
UX-165 Add padding and maxWidth system props to modal

jonambas:
### What Changed
- Adds `padding` and `maxWidth` system props to modal with defaults

### How To Test or Verify
- Verify max width and padding properties work

### PR Checklist


packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -68,7 +73,7 @@ function Modal(props) {
         aria-modal="true"
       >
         <StyledWrapper>
-          <ModalContent open={props.open} maxWidth={props.maxWidth}>
+          <ModalContent open={props.open} maxWidth={maxWidth}>
```
logansparlin, line 76: Any reason to not add `open` to the object of props above and just do `{open}`

PR #411, opened 2020-05-15
UX-161 Text Component Updates

logansparlin:
### What Changed
- remove crop functionality
- remove lodash
- adds looksLike prop

### How To Test or Verify
- `npm run start:storybook`
- `npm run test:unit`
### PR Checklist

- [ ] Update `unreleased.md` in the root directory


package.json
```diff
@@ -35,6 +35,7 @@
     "@styled-system/prop-types": "^5.1.2",
     "@styled-system/props": "^5.1.5",
     "downshift": "^3.2.10",
+    "eslint-plugin-lodash": "^2.7.0",
```
jonambas, line 38: This should be a dev dep right?

packages/matchbox/src/components/Text/styles.js
```diff
@@ -1,51 +1,69 @@
...
+      break;
+    case 'h6':
+      styles = `
+        font-size: ${tokens.fontSize_200};
+        line-height: ${tokens.lineHeight_200};
+        font-weight: ${tokens.fontWeight_semibold};
+      `;
+      break;
+    case 'p':
+    default:
```
jonambas, line 60: One comment actually. I don't think this should default to `p` right? so the styles still work correctly without the `looksLike` prop. Default should return nothing

packages/matchbox/src/components/Text/Text.js
```diff
@@ -1,28 +1,36 @@
...
 
-const system = compose(
-  color,
-  space,
-  typography
-);
+const system = compose(color, space, typography);
 
-const Text = styled('p')`
+const StyledText = styled(Box)`
```
jonambas, line 10: Caught this one too, I think this should stay `p`

packages/matchbox/src/components/Text/Text.js
```diff
@@ -1,28 +1,36 @@
...
   ${truncate}
-  ${crop}
+  ${looksLike}
 `;
 
+const Text = function(props) {
+  const { as, looksLike, children, ...rest } = props;
+
+  return (
+    <StyledText as={as} looksLike={looksLike} {...rest}>
```
jonambas, line 20: Yeah think this is whats causing the jest error, could just rename it before passing it through and maybe add a comment so we remember why we did it

`looks={looksLike}`

packages/matchbox/src/components/Text/tests/Text.test.js
```diff
@@ -20,20 +14,40 @@ describe('Text', () => {
     expect(wrapper).toMatchSnapshot();
```
jonambas, line 14: Screen Shot 2020-05-20 at 11 10 12 AM

Looks like the default isn't working correctly i think

PR #413, opened 2020-05-18
UX-155 Add Tab keyboard navigation

jonambas:
### What Changed
- Adds support for tab keyboard navigation
- Adds new prop `keyboardActivation` defaults to `auto`
- Adds new `useTabs` hook to support tab state implementation
- Refs can now be forwarded to `Text`, `UnstyledLink`
- Tabs with custom components now require a `ref` to be forwarded

Key | Desc
-- | --
Enter / Spacebar | Activates focused tab – inherited through button and anchor tags with href
Tab | When focus is before tablist moves focus to the active tab. If focus is on the active tab moves focus to the next element in the keyboard focus order, and skip the next tab.
ArrowLeft / ArrowRight | Navigates between tabs when orientation is "horizontal". When on last, should move to first. When on first, should move to last.
Home / PageUp | Navigates to the last tab in the TabList.
End / PageDown | Navigates to the first tab in the TabList.

### How To Test or Verify
- Run storybook `npm run start:storybook`
- Visit tab stories and verify key events work with both auto and manual activation

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Tabs/Tab.js
+++ b/packages/matchbox/src/components/Tabs/Tab.js
@@ -1,0 +1,51 @@
+import React from 'react';
+import styled from 'styled-components';
+import { UnstyledLink } from '../UnstyledLink';
+import { tabStyles } from './styles';
+
+// TODO Replace this when styled-components supports shouldForwardProps
+// See: https://github.com/styled-components/styled-components/commit/e02109e626ed117b76f220d0b9b926129655262d
+// Or when UnstyledLink is updated to use system props
+const LinkWrapper = React.forwardRef(function LinkWrapper(props, ref) {
+  const { selected, fitted, ...rest } = props;
+  return <UnstyledLink ref={ref} {...rest} />;
+});
+
+const StyledTab = styled(LinkWrapper)`
+  ${tabStyles}
+`;
+
+const Tab = React.forwardRef(function Tab(props, ref) {
+  const { index, content, selected, fitted, component, Component, tabIndex, ...rest } = props;
+
+  function handleClick(event) {
+    const { index, onClick } = props;
+    onClick(event, index);
+  }
+
+  // Buttons ensure focusability
+  // Links will be focusable with an href
+  // TODO deprecate `Component`
+  const wrapper = component || Component || 'button';
+
+  return (
+    <StyledTab
+      aria-selected={selected === index}
+      component={wrapper}
+      selected={selected === index}
+      fitted={fitted}
+      ref={ref}
+      {...rest}
+      onClick={handleClick}
+      role="tab"
+      tabIndex={tabIndex}
+      type="button"
+    >
+      {content}
+    </StyledTab>
+  );
+});
+
+Tab.displayName = 'Tab';
+
+export default Tab;
--- a/packages/matchbox/src/components/Tabs/Tabs.js
+++ b/packages/matchbox/src/components/Tabs/Tabs.js
@@ -28,11 +28,15 @@ const Container = styled('div')`
 `;
 
 function Tabs(props) {
-<<<<<<< HEAD
-  const { disableResponsiveBehavior, tabs, selected, onSelect, fitted, ...rest } = props;
-=======
-  const { tabs, selected, onSelect, fitted, keyboardActivation, ...rest } = props;
->>>>>>> 07501457... UX-155 Add keyboard navigation for tabs
+  const {
+    disableResponsiveBehavior,
+    keyboardActivation,
+    tabs,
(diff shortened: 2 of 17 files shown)


jest.config.json
```diff
@@ -4,19 +4,14 @@
...
     "packages/matchbox-icons/src/IconBase/**/*.{js,jsx}"
   ],
-  "coveragePathIgnorePatterns": [
-    "index.js"
-  ],
+  "coveragePathIgnorePatterns": ["index.js"],
   "coverageThreshold": {
     "global": {
-      "statements": 90
+      "statements": 85
```
jonambas, line 10: Only changed this in this file, rest is prettier

packages/matchbox/src/components/Tabs/Tabs.js
```diff
@@ -192,6 +152,7 @@ Tabs.propTypes = {
     PropTypes.oneOf(['orange', 'blue', 'navy', 'purple', 'red']),
     'Tab color is no longer configurable',
   ),
+  keyboardActivation: PropTypes.oneOf(['auto', 'manual']),
```
jonambas, line 155: New prop

packages/matchbox/src/components/Tabs/styles.js
```diff
@@ -8,9 +8,10 @@ export const tabStyles = ({ selected, fitted }) => `
   position: relative;
   flex: ${fitted ? '1' : '0'};
   text-decoration: none;
+  outline: none;
   
   padding: 0 ${tokens.spacing_200};
-  margin: 0 ${!fitted ? `${tokens.sizing_450}` : '0'};
+  margin: 0 ${!fitted ? `${tokens.sizing_450}` : `${tokens.sizing_200}`};
```
jonambas, line 14: I made this change so the focus outlines dont get cut off when fitted

packages/matchbox/src/helpers/keyEvents.js
```diff
@@ -2,57 +2,71 @@ const keys = {
...
+    key: 'PageDown',
+    keyCode: 34,
+    shiftKey: false,
+  },
 };
 
 function compareEvent(event, e) {
-  return (e.key === keys[event].key || e.keyCode === keys[event].keyCode) && e.shiftKey === keys[event].shiftKey;
+  return (
+    !!keys[event] &&
```
jonambas, line 66: added this check in case the event doesnt exist in the object above, and the two new key events, rest is prettier

packages/matchbox/src/components/Tabs/Tabs.js
```diff
@@ -38,39 +27,16 @@ const Container = styled('div')`
...
-      type="button"
-    >
-      {content}
-    </StyledTab>
-  );
-}
-
-Tab.displayName = 'Tab';
-
 function Tabs(props) {
```
jonambas, line 30: forwardRef

packages/matchbox/src/components/Text/Text.js
```diff
@@ -12,20 +12,20 @@ const StyledText = styled('p')`
...
     </StyledText>
   );
-};
+});
 
 Text.propTypes = {
   as: PropTypes.elementType.isRequired,
   children: PropTypes.node.isRequired,
-  lookslike: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p']),
+  looksLike: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'p']),
```
logansparlin, line 28: need to change the stories to use `looksLike`

packages/matchbox/src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import styled from 'styled-components';
+import { UnstyledLink } from '../UnstyledLink';
+import { tabStyles } from './styles';
+
+// TODO Replace this when styled-components supports shouldForwardProps
+// See: https://github.com/styled-components/styled-components/commit/e02109e626ed117b76f220d0b9b926129655262d
+// Or when UnstyledLink is updated to use system props
+const LinkWrapper = React.forwardRef(function LinkWrapper(props, ref) {
+  const { selected, fitted, ...rest } = props;
```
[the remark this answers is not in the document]
jonambas: Yeah this is here to omit the props from getting passed down, I think when we upgrade styled-components we can remove this

packages/matchbox/src/components/Text/Text.js
jonambas: Woops forgot I updated this

PR #415, opened 2020-05-20
UX-167 Spinner Component

logansparlin:
### What Changed
- Adds new Spinner Component to Matchbox
- Supports `color` `size` and `label` props

### How To Test or Verify
- Run `npm run start:storybook`
- Visit Spinner story 

### PR Checklist

- [ ] Update `unreleased.md` in the root directory


packages/matchbox/src/components/Spinner/styles.js
```diff
@@ -0,0 +1,87 @@
+import { tokens } from '@sparkpost/design-tokens';
+import { keyframes, css } from 'styled-components';
+
+export const circleOuter = props => {
+  let size = '60px';
+
+  if (props.size === 'small') {
+    size = '28px';
```
jonambas, line 8: I think small needs to be a bit smaller to fit inside small buttons, I also tried putting it into textfield as a prefix/suffix. Could also maybe make this a default size and just add another smaller one

```js
<Button size="small">
      <Spinner size="small" label="loading" />
    </Button>
<TextField prefix={<Spinner size="small" label="loading" />} />
```

packages/matchbox/src/components/Spinner/styles.js
```diff
@@ -0,0 +1,87 @@
...
+    animation: ${rotateAnimation} 2s linear infinite;
+  `;
+};
+
+export const circle = props => {
+  let color;
+
+  switch (props.color) {
+    case 'gray':
+      color = tokens.color_brand_gray;
```
jonambas, line 24: Should this be a gray token and not the brand gray?

packages/matchbox/src/components/Spinner/Spinner.js
```diff
@@ -0,0 +1,59 @@
...
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { margin, layout, compose } from 'styled-system';
+import { pick } from '@styled-system/props';
+import { createPropTypes } from '@styled-system/prop-types';
+import { Box } from '../Box';
+import { ScreenReaderOnly } from '../ScreenReaderOnly';
+import { circleOuter, circle } from './styles';
+
+const system = compose(margin, layout);
```
jonambas, line 11: Since `size` is also a system prop I see this in the css - I think all we need is height and width right? Do you think we still need this? I might have wrote the AC wrong in the ticket, I _think_ I meant `position` not `layout` 🤦 


Screen Shot 2020-05-26 at 6 56 08 PM

packages/matchbox/src/components/Spinner/styles.js
```diff
@@ -0,0 +1,127 @@
+import { tokens } from '@sparkpost/design-tokens';
+import { keyframes, css } from 'styled-components';
+
+export const circleOuter = props => {
+  let size;
+
+  switch (props.size) {
+    case 'small':
+      size = '20px';
```
jonambas, line 9: 👍 size looks good

packages/matchbox/src/components/Spinner/Spinner.js
logansparlin: Yea that makes more sense. Since we've got size defined in a prop we won't need to change that with the `layout` prop.

PR #424, opened 2020-05-26
UX-218 Design Feedback & Cleanup

jonambas:
### What Changed
- Set default margins/padding on Stack, Panel, and Tables to 500
- Table header and row height now matches mocks
- Links inside helptext are now gray
- Resized the modal close button
- Adds a prop to disable Tab responsive behavior for the templates  ui

### How To Test or Verify
- Run Storybook & verify the changes above

### PR Checklist


packages/matchbox/src/components/HelpText/HelpText.js
```diff
@@ -1,6 +1,15 @@
 import React from 'react';
-import { Box } from '../Box';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { tokens } from '@sparkpost/design-tokens';
+import { Box } from '../Box';
+
+const StyledText = styled(Box)`
```
logansparlin, line 7: Looks good but still rendering `Box` below, need to change to this `StyledText`

packages/matchbox/src/components/Tabs/Tabs.js
```diff
@@ -70,7 +70,7 @@ function Tab(props) {
 Tab.displayName = 'Tab';
 
 function Tabs(props) {
-  const { tabs, selected, onSelect, fitted, ...rest } = props;
+  const { disableResponsiveBehavior, tabs, selected, onSelect, fitted, ...rest } = props;
```
logansparlin, line 73: When responsive behavior is disabled and the page is sized down the tabs overflow off the side of the page. Not sure if this will ever be an issue or not, but would it make sense to hide the overflow or even give it a scrollable overflow?

jonambas: Hmm I think hiding makes sense. Nick requested this so not sure what he is expecting

PR #441, opened 2020-06-01
UX-214 Add keyboardkey component

jonambas:
### What Changed
- Adds new `KeyboardKey` component

Screen Shot 2020-06-01 at 2 35 05 PM

### How To Test or Verify
- Run storybook and visit the keyboarkey story

### PR Checklist


packages/matchbox/src/components/KeyboardKey/KeyboardKey.js
```diff
@@ -0,0 +1,36 @@
...
+  color: ${tokens.color_gray_900};
+  font-size: ${tokens.fontSize_100};
+  line-height: ${tokens.lineHeight_100};
+
+  box-shadow: 0px 3px 0px -1px ${tokens.color_gray_200}, 0px 3px 0px ${tokens.color_gray_400};
+`;
+
+const KeyboardKey = React.forwardRef(function KeyboardKey(props, ref) {
+  return (
+    <StyledKey as="kbd" ref={ref}>
```
logansparlin, line 25: do we need the `as="kbd"` here since it's already defined above?

packages/matchbox/src/components/KeyboardKey/KeyboardKey.js
```diff
@@ -0,0 +1,36 @@
...
+import { tokens } from '@sparkpost/design-tokens';
+
+const StyledKey = styled.kbd`
+  position: relative;
+  display: inline-block;
+  padding: 0 ${tokens.spacing_100};
+  height: ${tokens.sizing_450};
+
+  background: ${tokens.color_gray_100};
+  border-radius: 4px;
```
logansparlin, line 13: Is this something we want to add to the tokens? Probably not necessary if we're just using 4px border radius in this one instance

packages/matchbox/src/components/KeyboardKey/KeyboardKey.js
jonambas: Ah yea good catch

packages/matchbox/src/components/KeyboardKey/KeyboardKey.js
jonambas: I thought about it, but I don't _think_ well be using it anywhere else

PR #484, opened 2020-06-18
UX-239 Add react-scrolllock to modals and drawers

jonambas:
### What Changed
- Locks window scroll when drawers or modals are open

### How To Test or Verify
- Run storybook and open a drawer or modal with background content that overflows the iframe, test scrolling

### PR Checklist


packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -60,51 +61,62 @@ function Modal(props) {
...
+      </Portal>
+    </>
   );
-}
+});
 
-function ModalContent(props) {
+const ModalContent = React.forwardRef(function ModalContent(props, userRef) {
   const { open, children, maxWidth } = props;
-  const content = useRef(null);
```
jonambas, line 107: I found this ref wasn't being used at all

PR #490, opened 2020-06-19
Resolve Issue #473 Delay Tooltip

### LINKED ISSUE #473: Delay display of tooltips
## 🚀 Feature request

**Current Behavior**

Tooltips are displayed almost instantly when hovering.  In this example with checkboxes, the tooltip hides other checkbox options.

Screen Shot 2020-06-17 at 7 40 09 AM

**Suggested Solution**

Delay the display of a tooltip to match [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title).

All users.

jonambas:
### What Changed
- Closes #473 

Kapture 2020-06-19 at 10 10 32

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Tooltip/tests/Tooltip.test.js
+++ b/packages/matchbox/src/components/Tooltip/tests/Tooltip.test.js
@@ -11,10 +11,11 @@ describe('Tooltip', () => {
     );
 
   const container = wrapper => wrapper.find('div').at(2);
-  const content = wrapper => wrapper.find('div').at(3);
+  const content = wrapper => wrapper.find('div').at(4);
 
   it('should render default styles correctly', () => {
     const wrapper = subject();
+    wrapper.find('button').simulate('mouseOver');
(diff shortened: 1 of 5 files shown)


packages/matchbox/src/components/Tooltip/tests/Tooltip.test.js
```diff
@@ -81,6 +60,7 @@ describe('Tooltip', () => {
 
   it('should pass through system props', () => {
     const wrapper = subject({ bg: 'blue', fontSize: '400', pr: '500' });
+    wrapper.find('button').simulate('mouseOver');
```
logansparlin, line 63: Should we test for focus here, too, or just in the cypress tests?

jonambas: I don't think theres a way to test focus in jest, I think the cypress test is enough though

PR #509, opened 2020-06-26
Add border, width, height system props to Panel

jonambas:
### What Changed
- Adds border, width, height system props to the Panel component


stories/layout/Panel.stories.js
```diff
@@ -119,6 +119,38 @@ export const SystemProps = withInfo()(() => (
   </>
 ));
 
+export const AsCards = withInfo()(() => (
```
logansparlin, line 122: Should probably update the System Props story to include the new system props

PR #515, opened 2020-06-30
UX-246 Collapsible Nav Sections

logansparlin:
### PR Checklist


site/src/components/ExpandableMenuItem/ExpandableMenuItem.js
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+
+import { ArrowForward, ArrowDownward } from '@sparkpost/matchbox-icons';
```
jonambas, line 4: Do you think chevrons would be better than arrows here?

site/src/components/SideNavigation/SideNavigation.js
```diff
@@ -1,53 +1,112 @@
...
-    );
+  React.useEffect(function() {
+    setPosition(getRectFor(ref.current));
+  }, []);
+
+  function onScroll() {
+    let { scrollY } = window;
+
+    if (scrollY >= position.top) {
+      setFixed(true);
```
jonambas, line 86: I think you can do this with `position="sticky"` also instead of using javascript

site/src/components/ExpandableMenuItem/ExpandableMenuItem.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  function navigateToFirstLink() {
+    if (!defaultExpanded) {
+      navigate(firstRoute);
+    }
+  }
+
+  return (
+    <div onClick={navigateToFirstLink}>
```
jonambas, line 35: I think this needs `role="button"` or make it a `button` element

site/src/components/ExpandableMenuItem/ExpandableMenuItem.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const [expanded, setExpanded] = useState(defaultExpanded || false);
+
+  function toggleExpanded() {
+    setExpanded(!expanded);
+  }
+
+  function navigateToFirstLink() {
+    if (!defaultExpanded) {
+      navigate(firstRoute);
```
jonambas, line 30: This is good, but feels a little janky - i found this, I think it would be good to incorporate:
https://www.gatsbyjs.org/docs/browser-apis/#shouldUpdateScroll

in gatsby browser file:
```js
exports.shouldUpdateScroll = ({
  routerProps: { location },
  getSavedScrollPosition
}) => {

 // const currentPosition = getSavedScrollPosition(location);
 // Maybe we can do a smooth scroll to top here on route change?
  return false;
};
```

jonambas: I was thinking of implementing something to behave similar to the polaris site – it does a smooth scroll to the top on route change.  we'd need to test with #` anchor links too to make sure those still work

site/src/routes/routes.json
```diff
@@ -63,9 +63,10 @@
   {
     "path": "/components",
     "label": "Components",
+    "indexRoute": "/components/overview",
```
jonambas, line 66: Need to do the same with `/design` overview, but looks good!

PR #526, opened 2020-07-08
UX-274 Add default timeout prop to Button Transition

logansparlin:
### PR Checklist


packages/matchbox/src/components/Button/Button.js
```diff
@@ -111,7 +111,7 @@ const Button = React.forwardRef(function Button(props, ref) {
 
   const loadingIndicator = React.useMemo(() => {
     return (
-      <Transition mountOnEnter unmountOnExit in={loading}>
+      <Transition mountOnEnter unmountOnExit in={loading} timeout="500">
```
jonambas, line 114: I think this needs to be a number: https://reactcommunity.org/react-transition-group/transition#Transition-prop-timeout

And we should also just use 0 here for no timeout

PR #528, opened 2020-07-09
UX-250 Responsive Navigation

logansparlin:
### PR Checklist


site/src/components/SecondaryNavigation/SecondaryNavigation.js
```diff
@@ -0,0 +1,50 @@
...
+import { ChevronLeft, ChevronRight } from '@sparkpost/matchbox-icons';
+import SideNavigation from '../SideNavigation/SideNavigation';
+
+function SecondaryNavigation(props) {
+  const { navItems } = props;
+
+  const { label: primaryLabel, childRoutes } = _.find(navItems, [
+    'selected',
+    true
+  ]);
```
jonambas, line 14: I think the build error is coming from here – maybe provide a default if there are no navitems like `= _.find(navItems, ['selected', true]) || {}`

PR #563, opened 2020-07-24
UX-237 Add new Panel, Panel.Header, Panel.Section, Panel.Action components

### LINKED ISSUE #467: Need a way to handle Panel "header" UI with a border
## 🚀 Feature request

**Current Behavior**
`Panel` supports passing in a `title` prop, which handles most scenarios, however, there are certain cases in which having a border below the title (along with a little bit of padding) would be nice. Here is an example:

**Before:**
Screen Shot 2020-06-16 at 11 07 52 AM

**After:**
Screen Shot 2020-06-16 at 11 08 27 AM

I accomplished this by doing the following:

```js
<Panel title="Message History">
  <Box paddingTop="1.25rem" border="400"/> // `paddingTop` value is funky to adjust for button height inside - otherwise an alignment problem is introduced!
</Panel>
```
**Desired Behavior**

Ideally, this could be handled through compound component composition.

**Suggested Solution**
```js
<Panel>
  <Panel.Header borderBottom="whatever"> // defaults to no border, but styled-system props could be exposed here
    <Panel.Title>Message History</Panel.Title>
  </Panel.Header>
</Panel>
```

Makes FE development easier and easier to understand.

jonambas:
### What Changed
- Adds new `Panel` `Panel.Header` `Panel.Section` `Panel.Action` components
- Closes #467

### How To Test or Verify
- Visit stories and verify the new Panel component works as expected
- http://localhost:9001/?path=/story/layout-panel--with-a-header

### PR Checklist


packages/matchbox/src/components/Panel/Header.js
```diff
@@ -1,45 +1,49 @@
...
+    <Box
+      borderBottom={borderBottom ? '400' : 'none'}
+      className={className}
+      p="500"
+      pb={borderBottom ? null : '0'} // Null to inherit p="500"
+      {...paddingContext}
+      ref={userRef}
+      tabIndex="-1"
+    >
+      <Columns collapseBelow="xs" space="300">
```
logansparlin, line 26: Action buttons look a little off on mobile when collapsed because of button padding. Not sure if we want to do something about that or not. 
Screen Shot 2020-08-03 at 11 52 14 AM

packages/matchbox/src/components/Panel/Header.js
```diff
@@ -1,45 +1,49 @@
...
-
   return (
-    <HeaderOuter {...paddingContext} {...rest} pb="0">
-      <HeaderText>{title}</HeaderText>
-      {actionMarkup}
-    </HeaderOuter>
+    <Box
+      borderBottom={borderBottom ? '400' : 'none'}
+      className={className}
+      p="500"
```
logansparlin, line 20: 1rem padding on mobile?

packages/matchbox/src/components/Panel/Panel.js
```diff
@@ -1,104 +1,69 @@
...
+        border="400"
+        borderRadius="100"
+        className={className}
+        position="relative"
+        {...innerSystemProps}
+        height={innerHeight}
+      >
+        {accent && <Accent accentColor={accent} />}
+        <PanelPaddingContext.Provider
+          value={{ p: contextP || contextPadding || '500', ...context }}
```
logansparlin, line 40: Default to 1rem padding on mobile?

packages/matchbox/src/components/Panel/Header.js
logansparlin: Also, do we even need this if we're using paddingContext?

PR #564, opened 2020-07-27
UX-284 Darken button hover background colors

jonambas:
### What Changed
- Darkens background hover colors for `outline` `outlineBorder` and `flat` buttons

### How To Test or Verify
- Run storybook and visit the button color story

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Button/styles.js
+++ b/packages/matchbox/src/components/Button/styles.js
@@ -90,25 +90,25 @@ export const colorVariant = props => {
   switch (props.buttonColor) {
     case 'orange': // To be deprecated
     case 'blue':
-      color = tokens.color_blue_700;
-      darkHoverColor = tokens.color_blue_800;
-      lightActiveColor = tokens.color_blue_200;
-      lightHoverColor = tokens.color_blue_100;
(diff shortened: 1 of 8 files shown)


packages/matchbox/src/components/Button/styles.js
```diff
@@ -90,25 +90,25 @@ export const colorVariant = props => {
...
 
     case 'gray':
     default:
-      color = tokens.color_gray_900;
-      darkHoverColor = tokens.color_gray_1000;
-      lightActiveColor = tokens.color_gray_200;
-      lightHoverColor = tokens.color_gray_100;
+      color = props.theme.colors.gray['900'];
+      darkHoverColor = props.theme.colors.gray['1000'];
+      lightActiveColor = props.theme.colors.gray['400'];
```
logansparlin, line 110: Do blue and red also need to be 2 steps darker than they were before?

jonambas: I kind of just eyeballed it, it seemed a little too light at first. What do you think?

logansparlin: I think they're good how they are

PR #577, opened 2020-08-05
UX-236 Banner.Action Component

logansparlin:
### What Changed

- Adds Banner.Action Component
- Deprecates `action` and `actions` props (but still renders them if used

### How To Test or Verify

- `npm run start:storybook`
- Verify Banner stories

### PR Checklist


packages/matchbox/src/components/Banner/Action.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Button } from '../Button';
+
+const Action = React.forwardRef(function Action(props, userRef) {
+  return <Button mt="500" mr="400" {...props} ref={userRef} />;
```
jonambas, line 5: Hm should we decrease these margins on mobile?

stories/feedback/Banner.stories.js
```diff
@@ -27,15 +27,9 @@ export const SuccessBanner = withInfo()(() => (
 ));
```
jonambas, line 28: Should we add a story here with old `action` to make sure they still work?

PR #578, opened 2020-08-06
UX-286 Add indeterminate checkbox support

jonambas:
### What Changed
- Adds support for `indeterminate` checked state
- Specifies `defaultChecked` prop doesn't work with intermediate, input must be fully controlled

### How To Test or Verify
- Run storybook and visit http://localhost:9001/?path=/story/form-checkbox--indeterminate-group

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Checkbox/Checkbox.js
+++ b/packages/matchbox/src/components/Checkbox/Checkbox.js
@@ -11,12 +11,20 @@ import { pick } from '@styled-system/props';
 import { omit } from '../../helpers/systemProps';
 import { createPropTypes } from '@styled-system/prop-types';
 import { margin } from 'styled-system';
-import { StyledLabel, StyledCheck, StyledBox, StyledInput, Wrapper } from './styles';
+import {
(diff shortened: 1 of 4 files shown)


packages/matchbox/src/components/Checkbox/Checkbox.js
```diff
@@ -105,7 +129,11 @@ Checkbox.displayName = 'Checkbox';
 Checkbox.Group = Group;
 Checkbox.propTypes = {
   id: PropTypes.string.isRequired,
-  checked: PropTypes.bool,
+  checked: PropTypes.oneOf([true, false, 'indeterminate']),
+  /**
+   * 'indeterminate' does not work with defaultChecked
+   */
+  defaultChecked: PropTypes.oneOf([true, false]),
```
logansparlin, line 136: Any reason to do this instead of PropTypes.bool?

jonambas: Oh good catch haha

PR #581, opened 2020-08-10
UX-287 Limit Modal Content Height

logansparlin:
### What Changed

- Limit modal content height and overflow scroll

### How To Test or Verify

- Verify Modal stories in Storybook

### PR Checklist


packages/matchbox/src/components/Modal/Content.js
```diff
@@ -7,7 +7,7 @@ const Content = React.forwardRef(function Content(props, ref) {
   const { children } = props;
 
   return (
-    <Box data-id="modal-content" p="500" ref={ref}>
+    <Box data-id="modal-content" p="500" maxHeight="60vh" overflow="scroll" ref={ref}>
```
jonambas, line 10: I would change this to overflowY so the horizontal scrollbar are hidden if someone has them enabled

Screen Shot 2020-08-12 at 9 02 57 AM

packages/matchbox/src/components/Modal/Modal.js
```diff
@@ -94,7 +94,13 @@ const Modal = React.forwardRef(function Modal(props, ref) {
             role="dialog"
             aria-modal="true"
           >
-            <Box p={['400', null, '700']} size="100%" ref={overlayRef} data-id="modal-overlay">
+            <Box
+              p={['400', null, '700']}
+              pt={['5rem', null]}
```
jonambas, line 99: should we use a token?

logansparlin: Might have to create a new one. Our tokens jump from 4rem to 10rem and 4rem is too small and the modal is still hidden behind the app header, 10rem is pretty large but I guess we could use it. What do you think?

jonambas: Oh yea, i'd just leave it at 5rem then, doesn't make sense to make an `850` token at `5rem`

PR #585, opened 2020-08-12
UX-294 Make media package public

logansparlin:
### PR Checklist


packages/matchbox-media/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "4.0.13",
   "description": "Images and Videos for Matchbox",
   "main": "matchbox-media.js",
+  "publicConfig": {
```
jonambas, line 6: I think this is `publishConfig`

logansparlin: Wow my brain really isn't working today

PR #609, opened 2020-08-25
UX-288 Add useBreakpoint hook

jonambas:
### What Changed
- Initial goal was to show 1 or 2 months on the Datepicker depending on window resize
- After speaking with @PatrickSison , we realized that `initialMonth` would have to also change on window resize
- Instead of implementing hidden internal behavior to transform month, we decided to just export a helper hook so FE can use it on the front end

```js
function Component() {
  const breakpoint = useBreakpoint();
  
  if (breakpoint === 'sm') {
    // do something
  }

  if (breakpoint === 'lg') {
    // do something
  }

  return ...
}
```

### How To Test or Verify
- Visit http://localhost:9001/?path=/story/form-datepicker--basic-datepicker

### PR Checklist


stories/form/DatePicker.stories.js
```diff
@@ -24,15 +23,71 @@ const modifiers = {
...
-    selectedDays={selectedDays}
-    onDayClick={d => console.log('click', d)}
-    m="400"
-  />
-));
+export const BasicDatepicker = withInfo({
+  propTables: [DatePicker],
+  source: false,
+  text: `
+  ~~~jsx
```
logansparlin, line 30: Interesting, this is cool

PR #610, opened 2020-08-26
UX-324 Page and Actionlist actions render with null actions

jonambas:
### What Changed
- Allows the `filterByVisible` function to fail gracefully by returning an empty array

### How To Test or Verify
- Run storybook
- Visit page stories: http://localhost:9001/?path=/story/layout-page--basic-example
- Try to add `null` or `undefined` into `secondaryActions`
- Do the same with actions here: http://localhost:9001/?path=/story/action-actionlist--deprecated-usage

### PR Checklist

### DIFF
--- a/packages/matchbox/src/helpers/array.js
+++ b/packages/matchbox/src/helpers/array.js
@@ -15,10 +15,11 @@
  *   2: [ { name: 'b', section: 2 } ]
  * }
  */
-export const groupBy = (list, groupingKey) => list.reduce((acc, item) => {
-  const index = item[groupingKey];
-  return { ...acc, [index]: [ ...(acc[index] || []), item ]};
-}, {});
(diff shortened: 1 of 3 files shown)


packages/matchbox/src/helpers/array.js
```diff
@@ -32,9 +33,15 @@ export const groupByValues = (list, groupingKey) => Object.values(groupBy(list,
  * // returns
  * [{ name: 'a' }, { name: 'c' }]
  */
-export const filterByVisible = (actions) => actions.reduce((acc, { visible = true, ...action }) => {
-  if (visible) {
-    acc.push(action);
+export const filterByVisible = actions => {
+  if (!actions) {
```
jonambas, line 37: Rest is prettier, just added this

PR #615, opened 2020-09-01
UX-289 Set Pagination pageRange to 3 on mobile

logansparlin:
### What Changed

- Use useBreakpoint hook to set Pagination pageRange to 3 on mobile

### How To Test or Verify

- `npm run start:storybook`
- Verify Pagination stories on mobile

### PR Checklist


jonambas (changes requested):
This looks good, we should write cypress tests to check for the behavior though. You could take a look at the datepicker test which shows how to change browser size in a test

jonambas (approved):
LGTM

PR #626, opened 2020-09-09
Resolves #625 - Add support for the className prop for the Column component

### LINKED ISSUE #625: Column component cannot be re-styled using `<Box/>` or `styled-components`
## 🐛 Bug report

**Current Behavior**
```
<Box as={Column} display={['none', 'block']} />
```

Styles are not passed on to the `Column` correctly. Similarly, overriding does not work when using `styled-components`:

```
const StyledColumn = styled(Column)`
  display: none;
`
```

**Expected behavior**

`<Box/>` and `styled-components` should work with all Matchbox components.

**Reproducible example**

See above example.

**Suggested solution(s)**

I believe the `className` prop just needs to be exposed on the component for this to work.

nicklemmon:
Resolves #625 

### What Changed

- `className` accepted as a prop for the `<Column/>` component

### How To Test or Verify

- On the Storybook site, try:
```js
<Box as={Column} display="none">I am invisible!</Box>
```

### PR Checklist

- [ ] Incorporate feedback

### DIFF
--- a/packages/matchbox/src/components/Column/Column.js
+++ b/packages/matchbox/src/components/Column/Column.js
@@ -14,7 +14,7 @@ const StyledColumn = styled(Box)`
 `;
 
 const Column = React.forwardRef(function Column(props, ref) {
-  const { width, children } = props;
+  const { width, children, className } = props;
   const { space, collapsed } = React.useContext(ColumnsContext);
 
   let columnWidth = width;
@@ -29,6 +29,7 @@ const Column = React.forwardRef(function Column(props, ref) {
 
   return (
     <StyledColumn
+      className={className}
       width={columnWidth}
       flex={!width && !collapsed ? '1' : ''}
       pt={collapsed ? space : null}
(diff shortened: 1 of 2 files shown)


packages/matchbox/src/components/Column/Column.js
```diff
@@ -46,6 +47,7 @@ Column.displayName = 'Column';
 Column.propTypes = {
   children: PropTypes.node,
   width: PropTypes.oneOfType([PropTypes.oneOf(['content']), PropTypes.number]),
+  className: PropTypes.string,
```
jonambas, line 50: I think this works. For your use case you could expose `display`

```jsx
import { display } from 'styled-system`;
import { createPropTypes } from '@styled-system/prop-types';

...

const { width, children, className, display } = props;

...

<StyledColumn display={display} />

...

Column.propTypes = {
  ...createPropTypes(display.propNames),
}

```

jonambas (approved):
Looks good, but added a suggestion

packages/matchbox/src/components/Column/Column.js
jonambas: `styled(Box)` already includes `display` so no need to add it to the `StyledColumn` declaration

jonambas (approved):
LGTM

PR #643, opened 2020-09-21
UX-331 Add LabelValue component

jonambas:
### What Changed
- Adds new component `LabelValue`

### How To Test or Verify
- Visit LabelValue stories: http://localhost:9001/?path=/story/layout-labelvalue--vertical-orientation

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/LabelValue/LabelValue.js
+++ b/packages/matchbox/src/components/LabelValue/LabelValue.js
@@ -16,7 +16,7 @@ const LabelValue = React.forwardRef(function LabelValue(props, userRef) {
   const systemProps = pick(rest, margin.propNames);
 
   return (
-    <StyledWrapper className={className} {...systemProps} ref={userRef}>
+    <StyledWrapper className={className} {...systemProps} ref={userRef} tabIndex="-1">
       <Box
         display={orientation === 'horizontal' ? 'grid' : 'block'}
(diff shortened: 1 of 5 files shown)


packages/matchbox/src/components/LabelValue/LabelValue.js
```diff
@@ -0,0 +1,48 @@
...
+  const { label, children, className, orientation, ...rest } = props;
+
+  const systemProps = pick(rest, margin.propNames);
+
+  return (
+    <StyledWrapper className={className} {...systemProps} ref={userRef} tabIndex="-1">
+      <Box
+        display={orientation === 'horizontal' ? 'grid' : 'block'}
+        gridGap={orientation === 'horizontal' ? '300' : ''}
+        gridTemplateColumns={orientation === 'horizontal' ? `${tokens.spacing_900} 1fr` : ''}
```
logansparlin, line 24: Are we always going to want this much horizontal spacing between the label and the value? Do you think we'd want to make this configurable or give it a max width?

logansparlin (approved):
This looks great. One small comment, but if you're okay with it then it's good to merge

packages/matchbox/src/components/LabelValue/LabelValue.js
jonambas: In the app, the component uses `11rem` which is larger I think, at 176px

PR #647, opened 2020-09-22
UX-342 Support  prop on Modal.Content

jonambas:
### What Changed
- Removes modal content default prop `p`
- Hardcodes the default instead so `padding` works when passed to the component

### How To Test or Verify
- Visit the modal story: http://localhost:9001/?path=/story/overlays-modal--system-props
- Verify `p` works as well as `padding`

### PR Checklist


logansparlin (approved):
lgtm

PR #652, opened 2020-09-29
UX-335 Add ComboBoxTextField delimiter prop

jonambas:
### What Changed
- Adds `delimiter` string prop to the combobox textfield

### How To Test or Verify
- Visit http://localhost:9001/?path=/story/form-combobox--with-delimiter
- Verify delimiter works

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/ComboBox/ComboBoxTextField.js
+++ b/packages/matchbox/src/components/ComboBox/ComboBoxTextField.js
@@ -139,7 +139,7 @@ ComboBoxTextField.propTypes = {
   autoFocus: PropTypes.bool,
   children: PropTypes.node,
   disabled: PropTypes.bool,
-  delimiter: PropTypes.node,
+  delimiter: PropTypes.string,
   error: PropTypes.string,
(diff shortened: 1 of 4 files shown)


packages/matchbox/src/components/ComboBox/ComboBoxTextField.js
```diff
@@ -85,11 +86,25 @@ function ComboBoxTextField(props) {
...
+                  <span key={i}>
+                    <Tag onRemove={!disabled ? () => removeItem(item) : null}>
+                      {itemToString(item)}
+                    </Tag>
+                    {delimiter && i < selectedItems.length - 1 ? (
+                      <Box
+                        as="span"
+                        ml="100"
+                        fontWeight="medium"
+                        style={{ textTransform: 'uppercase', fontStyle: 'italic' }}
```
logansparlin, line 100: Doesn't really matter either way, but `fontStyle` is a prop available on `Box` so could do that instead of assigning it here

logansparlin (approved):
Looks good!

PR #653, opened 2020-09-29
UX-339 Componentize EmptyState

logansparlin:
### What Changed

- Componentized `EmptyState`
- Matched `EmptyState` styles to mocks https://sparkpost.invisionapp.com/overview/Empty-States-ckcno05yk00tf011ij29ebpmx/screens
- Moved old `EmptyState` to `EmptyState.LEGACY`

### How To Test or Verify

- `npm run start:storybook`
- Verify EmptyState stories

### PR Checklist


jonambas:
Screen Shot 2020-10-02 at 11 07 15 AM

^ looks like support-docs got added as a sub module

packages/matchbox/src/components/EmptyState/Image.js
```diff
@@ -0,0 +1,31 @@
...
+import { Picture } from '../Picture';
+
+const Image = React.forwardRef(function Image(props, userRef) {
+  const { src } = props;
+
+  return (
+    <StyledImage
+      display={['none', null, 'block']}
+      width="100%"
+      maxWidth="600px"
```
jonambas, line 14: This feels a _bit_ small to be, maybe because of my monitor. Maybe we could make this larger based on breakpoint?

jonambas: Or maybe it should just expand to the 50% container width?

packages/matchbox/src/components/EmptyState/Image.js
```diff
@@ -0,0 +1,31 @@
...
+      height="auto"
+      ref={userRef}
+    >
+      <Picture>
+        <Picture.Image src={src} />
```
jonambas, line 19: I would pass through `children` as well here, before `Picture.Image` in case we need to pass additional sources here

jonambas: Might also be a good idea to pass through `className`

packages/matchbox/src/components/EmptyState/styles.js
```diff
@@ -6,13 +6,23 @@ export const StyledContent = styled(Box)`
   p {
     color: ${props => props.theme.colors.gray['700']};
```
jonambas, line 7: I think the color is a little darker in the mocks

jonambas (changes requested):
Looks pretty good, some small comments:

packages/matchbox/src/components/Page/Page.js
```diff
@@ -139,9 +139,9 @@ function Page(props) {
 
   if (show) {
     return (
-      <EmptyState title={title} primaryAction={primaryAction} {...emptyProps}>
+      <EmptyState.LEGACY title={title} primaryAction={primaryAction} {...emptyProps}>
         {content}
-      </EmptyState>
+      </EmptyState.LEGACY>
     );
   }
```
jonambas, line 147: One more change: can we `deprecate` the `empty` prop on Page

jonambas (approved):

PR #669, opened 2020-10-21
UX-351 create useInView hook

logansparlin:
### What Changed

- New `useInView` hook

### How To Test or Verify

- `npm run start:storybook`
- Verify `useInView` story

### PR Checklist


jonambas (changes requested):
This looks good, I think we need to add some tests for this. I think you can use Cypress to scroll the window with `cy.scrollTo(x, y)`

jonambas (approved):
looks good

PR #673, opened 2020-10-26
UX-357 Use resize observer for suffix and prefix measurements

jonambas:
### What Changed
- Uses `useResizeObserver` to set `prefix` and `suffix` widths in TextField

### How To Test or Verify
- Run storybook and visit http://localhost:9001/?path=/story/form-textfield--prefix-and-suffix
- Verify the widths are set correctly and horizontal padding is added to the input
- Note: I could not reproduce the bug referred to in the ticket

### PR Checklist


logansparlin (approved):
lgtm

PR #681, opened 2020-11-06
UX-367 Fix ListBox key events bug

logansparlin:
### What Changed

- Fixes a bug where using a number value on a `ListBox.Option` throws an error
- Adds `start:localMatchbox` script to the site to make testing matchbox changes in the design site easier

### How To Test or Verify

- `npm run start:storybook`
- Verify [ListBox stories](http://localhost:9001/?path=/story/form-listbox--basic-listbox)

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/ListBox/useOptionConstructor.js
+++ b/packages/matchbox/src/components/ListBox/useOptionConstructor.js
@@ -38,17 +38,17 @@ function useOptionConstructor({ options, value, onSelect, open, placeholder }) {
   }
 
   function clearKeysSoFarAfterDelay() {
+    let timeout = setTimeout(function() {
+      setKeysSoFar('');
+      setKeyClear(false);
+    }, 700);
+
     if (keyClear) {
       clearTimeout(timeout);
       setKeyClear(null);
     }
 
     setKeyClear(true);
-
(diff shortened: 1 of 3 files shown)


packages/matchbox/src/components/ListBox/useOptionConstructor.js
```diff
@@ -67,15 +67,15 @@ function useOptionConstructor({ options, value, onSelect, open, placeholder }) {
   }, [focused]);
 
   React.useLayoutEffect(() => {
-    if (keysSoFar) {
+    if (keysSoFar && options) {
       const index = options.findIndex(option =>
-        option.props.value.toLowerCase().startsWith(keysSoFar),
+        option.props.children.toLowerCase().startsWith(keysSoFar),
```
jonambas, line 72: Just to clarify, this is searching children (the option label) instead of the value, right?

jonambas (approved):
This looks good - was able to reproduce the bug and verify it is fixed

packages/matchbox/src/components/ListBox/useOptionConstructor.js
logansparlin: Yeah

PR #711, opened 2020-12-08
UX-402 Fix Listbox padding as inline element

jonambas:
### What Changed
- Adds additional padding-right to the ListBox component

### How To Test or Verify
- Run storybook, visit the new ListBox story: http://localhost:9001/?path=/story/form-listbox--as-an-inline-element

### PR Checklist


logansparlin (approved):
Looks good

PR #718, opened 2020-12-10
UX-403 & Issue-710 Fix empty Modal.Footer & dynamic children

### LINKED ISSUE #710: Modal.Footer should render passed in child components without restrictions
## 🚀 Feature request

**Current Behavior**

`Modal.Footer` only accepts `Button` components as children. In our app, we need to render links (usually that drive the router rather than true links) that _look like_ buttons but are semantically anchor tags:

**Desired Behavior**

The following should render successfully:
```js
<Modal.Footer>
  <PageLink as={Button}>Confirm Something</PageLink>
</Modal.Footer>
```

Unfortunately I think the only solution is to be less restrictive - I think as we have experimented with compound components in the app, that less restriction is ultimately better when it comes to handling children. I think this sort of a restrictive approach was something I had endorsed previously but now I'm not so sure.

See above.

Impacts FE engineers only, or may impact end users if a developer uses a non-ideal semantic element due to a component API restriction.

jonambas:
Closes #710 

### What Changed

- Modal.Footer buttons now use `variant` props instead of deprecated props
- Modal.Footer does not crash without children
- Modal.Footer now accepts non-Buttons as children, and renders them

### How To Test or Verify
- Run storybook and visit the new Modal stories:
- http://localhost:9001/?path=/story/overlays-modal--non-buttons
- http://localhost:9001/?path=/story/overlays-modal--empty-footer

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Modal/Footer.js
+++ b/packages/matchbox/src/components/Modal/Footer.js
@@ -1,22 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { getChild } from '../../helpers/children';
 import { Panel } from '../Panel';
 import { Box } from '../Box';
+import { Inline } from '../Inline';
 
 const buttonProps = {
   0: {
     color: 'blue',
-    mr: '400',
+    key: 'modal-primary',
   },
   1: {
     color: 'blue',
-    outlineBorder: true,
-    mr: '400',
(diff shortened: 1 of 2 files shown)


logansparlin (approved):
Looks good, using Inline was a good idea

PR #728, opened 2021-01-04
UX-385 Add ActionList usage and new Content section

jonambas:
### PR Checklist

### DIFF
--- a/site/src/pages/design/actionable-language.mdx
+++ b/site/src/pages/design/actionable-language.mdx
@@ -1,0 +1,40 @@
+---
+title: Actionable Language
+category: Content
+---
+
+# Actionable Language
+
+---
+
+### Buttons and Actions
+
+- Button and Action labels should be clear and concise.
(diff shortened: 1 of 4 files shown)


site/src/pages/design/actionable-language.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: Actionable Language
+category: Content
+---
+
+# Actionable Language
```
logansparlin, line 6: Nice. I think this could generally use a redesign at some point to make it a little more intuitive where to find the usage guides vs pattern guides but this is a nice flow for now

logansparlin (approved):
Just one comment on design that maybe we can think about for the future, but looks good!

site/src/pages/design/actionable-language.mdx
jonambas: Yeah, wasn't really sure where to put this stuff, but i think it works

PR #742, opened 2021-01-25
UX-411 Button should accept "as" and deprecate "component"

logansparlin:
### What Changed

- new `as` prop which replaces `component` and `Component`
- `component` and `Component` props are deprecated 
- updates tests to use `as`

### How To Test or Verify

- `npm run start:storybook`
- verify `As` button story

### PR Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24102,7 +24102,7 @@
         },
         "onetime": {
           "version": "1.1.0",
-          "resolved": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
+          "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
           "integrity": "sha1-ofeDj4MUxRbwXs78vEzP4EtO14k=",
           "dev": true
         },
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -24102,7 +24102,7 @@
         },
         "onetime": {
           "version": "1.1.0",
-          "resolved": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
+          "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
```
jonambas, line 24105: Not sure what happened here, can we restore this file?

packages/matchbox/src/components/Button/Button.js
```diff
@@ -206,9 +211,10 @@ Button.Group = Group;
 Button.Icon = Icon;
 
 Button.propTypes = {
+  as: PropTypes.elementType,
   children: PropTypes.node,
   color: PropTypes.oneOf(['gray', 'orange', 'blue', 'navy', 'purple', 'red', 'white']),
-  component: PropTypes.elementType,
+  component: deprecate(PropTypes.elementType, 'Use `as` instead'),
```
jonambas, line 217: On line 30, we should update that message too

jonambas (changes requested):
Couple small comments

Just realized, will have to do the same with UnstyledLink too, will cut a ticket for that

jonambas:
Oh also, should we update the documentation site in this PR too?

jonambas (approved):

PR #746, opened 2021-01-28
UX-431 Align chevron in Page component breadcrumbs

logansparlin:
### What Changed

- Fixes bug where chevron is not aligned with breadcrumbs in `Page` component

### How To Test or Verify

- `npm run start:storybook`
- Verify stories at http://localhost:9001/?path=/story/layout-page--basic-example

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Page/Page.js
+++ b/packages/matchbox/src/components/Page/Page.js
@@ -11,6 +11,7 @@ import { Popover } from '../Popover';
 import { UnstyledLink } from '../UnstyledLink';
 import { ScreenReaderOnly } from '../ScreenReaderOnly';
 import { filterByVisible } from '../../helpers/array';
+import { tokens } from '@sparkpost/design-tokens';
 
 const StyledLink = styled(UnstyledLink)`
   text-decoration: none;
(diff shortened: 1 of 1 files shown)


packages/matchbox/src/components/Page/Page.js
```diff
@@ -11,16 +11,17 @@ import { Popover } from '../Popover';
...
 const StyledLink = styled(UnstyledLink)`
   text-decoration: none;
 `;
 
 function Breadcrumb({ content, ...rest }) {
   return (
-    <StyledLink {...rest} fontSize="200" lineHeight="200" fontWeight="medium">
-      <Box as="span" display="inline-flex" align-items="center">
-        <ChevronLeft size={20} />
+    <StyledLink {...rest} fontSize="200" lineHeight="450" fontWeight="medium">
```
jonambas, line 22: Found a bug here, UnstyledLink doesn't accept `fontSize="200" lineHeight="450" fontWeight="medium"` at all, I don't see the styles being applied.

packages/matchbox/src/components/Page/Page.js
```diff
@@ -11,16 +11,17 @@ import { Popover } from '../Popover';
...
+      <Box as="span" display="inline-flex" align-items="center" lineHeight={tokens.spacing_450}>
```
jonambas, line 23: What's weird is storybook is applying this lineHeight, but in the App, this lineHeight is missing, which is causing the misalignment.

Not really sure what the fix is here

jonambas: Take a look here for an example: https://matchbox-storybook.netlify.app/?path=/story/layout-page--basic-example

jonambas: I _think_ all we need to do is make `align-items` -> `alignItems`? But still not sure why the bug isn't happening locally

packages/matchbox/src/components/Page/Page.js
jonambas: At least here it doesn't seem like it works: https://matchbox-storybook.netlify.app/?path=/story/layout-page--basic-example

jonambas (approved):
Verified changes through chrome dev tools – should update `alignItems` though

jonambas (approved):

PR #751, opened 2021-02-02
UX-432 Add aria-invalid to Checkbox Radio Select Textfield and ListBox

jonambas:
### What Changed
- Adds `aria-invalid` on all `<input>` components that support errors
- Skips `Toggle` and `RadioCard`

### How To Test or Verify
- Run storybook
- Verify `aria-invalid` is properly set on:
  - Checkbox
  - Radio 
  - Select
  - TextField
  - ListBox

### PR Checklist

### DIFF
--- a/cypress/integration/TextField.spec.js
+++ b/cypress/integration/TextField.spec.js
@@ -1,0 +1,14 @@
+describe('The TextField component', () => {
+  it('focuses on the `<input/>` element when clicking on the `<label>`', () => {
+    cy.visit('/iframe.html?id=form-textfield--basic-textfield');
+    cy.get('label').click();
+    cy.get('input').should('have.focus');
+    cy.get('[aria-invalid="false"]').should('exist');
+  });
+
+  it('renders with an error correctly', () => {
+    cy.visit('/iframe.html?id=form-textfield--with-an-error');
+    cy.get('[aria-invalid="true"]').should('exist');
+    cy.findAllByText('You forgot an ID!').should('exist');
+  });
+});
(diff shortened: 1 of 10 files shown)


packages/matchbox/src/components/TextField/TextField.js
```diff
@@ -138,6 +141,7 @@ const TextField = React.forwardRef(function TextField(props, userRef) {
     height: multiline ? 'auto' : '2.5rem',
     lineHeight: multiline ? '300' : '2.5rem',
     py: multiline ? '300' : '0',
+    hasError: !!error,
```
jonambas, line 144: Found a bug, we were never passing through `hasError` so the textfields were never red

packages/matchbox/src/components/ListBox/ListBox.js
```diff
@@ -202,7 +202,6 @@ const ListBox = React.forwardRef(function ListBox(props, userRef) {
               aria-haspopup="listbox"
               aria-pressed={open}
               aria-expanded={open}
-              aria-labelledby={id}
```
jonambas, line 201: Found another bug

cypress/integration/TextField.spec.js
```diff
@@ -0,0 +1,14 @@
...
+    cy.visit('/iframe.html?id=form-textfield--basic-textfield');
+    cy.get('label').click();
+    cy.get('input').should('have.focus');
+    cy.get('[aria-invalid="false"]').should('exist');
+  });
+
+  it('renders with an error correctly', () => {
+    cy.visit('/iframe.html?id=form-textfield--with-an-error');
+    cy.get('[aria-invalid="true"]').should('exist');
+    cy.findAllByText('You forgot an ID!').should('exist');
```
jonambas, line 12: Using `findAllByText` because there are multiple instances – the story source display also contains the string

logansparlin (approved):
Looks good, good catch on the error borders

PR #753, opened 2021-02-03
UX-439 Add data-id to Error component

jonambas:
### What Changed
- Hardcodes `data-id="error-message"` to the `Error` component
  - Surfaced in: TextField, Checkbox, Select, ListBox

### How To Test or Verify
- Run storybook
- Verify the attribute is set correctly on the error message

### PR Checklist


logansparlin (approved):
Looks good! For another ticket but realizing we should probably add cypress tests for radio at some point

PR #759, opened 2021-02-09
UX-446 Change <Page> action spacing from 500 to 300

logansparlin:
### What Changed

- `Page` actions now have `300` spacing instead of `500` to comply with patterns

### How To Test or Verify

- `npm run start:storybook`
- Verify Page [stories](http://localhost:9001/?path=/story/layout-page--basic-example) 

### PR Checklist


packages/matchbox/src/components/Snackbar/Snackbar.js
```diff
@@ -69,11 +69,6 @@ Snackbar.propTypes = {
    */
   status: PropTypes.oneOf(['default', 'success', 'danger', 'error', 'warning']),
 
-  /**
-   * Snackbar max-width in pixels.
-   */
-  maxWidth: PropTypes.number,
```
jonambas, line 75: Looks like you need to drop this commit

jonambas (approved):
Looks good, looks like the snackbar commit is in this PR too, just need to make sure to merge that one first so we have a clean history

PR #762, opened 2021-02-17
UX-445 Expandable update vertical padding with subtitle

logansparlin:
update vertical padding on Expandable with Subtitle

### What Changed

- updated vertical padding of `Expandable` with `subtitle` to match horizontal spacing

### How To Test or Verify

- `npm run start:storybook`
- Verify story http://localhost:9001/?path=/story/layout-expandable--with-image-title-and-subtitle

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Expandable/Expandable.js
+++ b/packages/matchbox/src/components/Expandable/Expandable.js
@@ -105,6 +105,7 @@ function Expandable(props) {
           ref={header}
           data-id="expandable-toggle"
           type="button"
+          variant={variant}
         >
           {iconMarkup}
           <Box display="inline-block" flex="1">
--- a/packages/matchbox/src/components/Expandable/Expandable.js
+++ b/packages/matchbox/src/components/Expandable/Expandable.js
@@ -23,7 +23,7 @@ const StyledTitle = styled('h3')`
   ${title}
 `;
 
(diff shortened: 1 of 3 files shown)


packages/matchbox/src/components/Expandable/styles.js
```diff
@@ -10,7 +10,7 @@ export const StyledHeader = styled('button')`
   ${() =>
     css({
       px: '450',
-      py: '200',
+      py: '300',
```
jonambas, line 13: I think we need to check if `variant === 'borderless'` here, and use `200` for borderless, `300` for regular ones

packages/matchbox/src/components/Expandable/Expandable.js
```diff
@@ -85,7 +85,11 @@ function Expandable(props) {
     </StyledIcon>
   ) : null;
 
-  const subtitleMarkup = subtitle ? <StyledSubtitle>{subtitle}</StyledSubtitle> : null;
+  const subtitleMarkup = subtitle ? (
+    <StyledSubtitle as="h6" mb="100">
```
jonambas, line 89: 👍

jonambas (changes requested):
Just one small change needed

jonambas (approved):
lgtm

PR #763, opened 2021-02-17
UX-394 Remove spreading of unknown ProgressBar props

jonambas:
### What Changed
- Removes spreading of unknown props on ProgressBar

### How To Test or Verify
- Verify ProgressBar story renders as expected: http://localhost:9001/?path=/story/feedback-progressbar--default
- Verify ProgressBar tests did not break

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/ProgressBar/ProgressBar.js
+++ b/packages/matchbox/src/components/ProgressBar/ProgressBar.js
@@ -23,7 +23,7 @@ export const StyledProgressBarInner = styled(Box)`
   ${calculatedWidth}
 `;
 
-function ProgressBar(props) {
+const ProgressBar = React.forwardRef(function ProgressBar(props, userRef) {
   const { completed = 0, label, size, valueText, ...rest } = props;
   const systemProps = pick(rest, system.propNames);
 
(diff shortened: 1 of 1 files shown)


packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -24,18 +25,19 @@ export const StyledProgressBarInner = styled(Box)`
...
       as="div"
       visualSize={size}
-      {...rest}
       role="progressbar"
       aria-label={label}
       aria-valuenow={`${completed}`}
       aria-valuemin="0"
       aria-valuemax="100"
       aria-valuetext={valueText}
+      {...systemProps}
```
jonambas, line 40: I think we should support data-id even if the UI doesn't currently use it

packages/matchbox/src/components/ProgressBar/ProgressBar.js
```diff
@@ -22,25 +23,28 @@ export const StyledProgressBarInner = styled(Box)`
...
+      data-id={rest['data-id']}
```
logansparlin, line 41: Any reason to do this over just explicitly defining the `data-id` prop in the spread?

logansparlin (approved):
Looks good, confirmed tests pass and FE not passing undefined props. Just one small comment

packages/matchbox/src/components/ProgressBar/ProgressBar.js
jonambas: Oh yea forgot you could do that

PR #764, opened 2021-02-17
UX-430 Create inverted color scheme for Panel component

logansparlin:
### What Changed

- New `appearance` prop on `Panel` component which accepts `inverted` or `default`
- Implemented with context to pass down to `Panel.Header` `Panel.SubHeader` and `Panel.Section`

### How To Test or Verify

- `npm run start:storybook`
- Verify `Panel` stories http://localhost:9001/?path=/story/layout-panel--inverted-appearance

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Panel/Panel.js
+++ b/packages/matchbox/src/components/Panel/Panel.js
@@ -36,7 +36,6 @@ const Panel = React.forwardRef(function Panel(props, userRef) {
         {...innerSystemProps}
         height={innerHeight}
         bg={appearance === 'inverted' ? 'gray.900' : ''}
-        borderColor={appearance === 'inverted' ? 'gray.600' : ''}
       >
         {accent && <Accent accentColor={accent} />}
         <PanelPaddingContext.Provider
--- a/packages/matchbox/src/components/Panel/Panel.js
+++ b/packages/matchbox/src/components/Panel/Panel.js
@@ -36,6 +36,7 @@ const Panel = React.forwardRef(function Panel(props, userRef) {
         {...innerSystemProps}
         height={innerHeight}
         bg={appearance === 'inverted' ? 'gray.900' : ''}
+        borderColor={appearance === 'inverted' ? 'gray.600' : ''}
       >
         {accent && <Accent accentColor={accent} />}
         <PanelPaddingContext.Provider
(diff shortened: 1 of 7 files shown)


packages/matchbox/src/components/Panel/Panel.js
```diff
@@ -35,12 +35,15 @@ const Panel = React.forwardRef(function Panel(props, userRef) {
         position="relative"
         {...innerSystemProps}
         height={innerHeight}
+        bg={appearance === 'inverted' ? 'gray.900' : ''}
```
jonambas, line 38: Need to change `border` as well, probably need to keep the border, but change borderColor?

packages/matchbox/src/components/Panel/SubHeader.js
```diff
@@ -17,7 +18,7 @@ const SubHeader = React.forwardRef(function SubHeader(props, userRef) {
       fontSize="200"
       fontWeight="normal"
       lineHeight="200"
-      color="gray.700"
+      color={appearanceContext === 'inverted' ? 'white' : 'gray.700'}
```
jonambas, line 21: SubHeader is typically a little muted, should we use a darker gray when its inverted?

packages/matchbox/src/components/Panel/Section.js
```diff
@@ -22,6 +22,7 @@ const Section = React.forwardRef(function Section(props, userRef) {
   const actions = getChild('Panel.Action', children);
   const content = excludeChild(['Panel.Action'], children);
   const paddingContext = React.useContext(PanelPaddingContext);
+  const appearanceContext = React.useContext(PanelAppearanceContext);
```
jonambas, line 25: Context is nice, but what happens if the user wants one Section normal, and another one inverted in the same Panel?

jonambas (changes requested):
I think we need to support the prop at the Section level too, what do you think?

packages/matchbox/src/components/Panel/Section.js
jonambas: Should we even support this use case though?

logansparlin: Would be easy enough to pass through to specific sections. Is our use case for these appearance changes a dark mode or something else, because if it's a dark mode I could see it being annoying having to specify an inverted appearance on all panel sections

jonambas: I think this is the main use case for now. 
Screen Shot 2021-02-24 at 5 19 28 PM

With dark mode enabled, I imagine "inverted" would probably stay dark, although we should talk about that when the time comes.

jonambas (approved):
This looks good. If we get around to visual testing we should def include these stories, so we don't have to manually test css styles in Jest or Cypress

PR #765, opened 2021-02-18
UX-397 Remove spreading of unknown Snackbar props

jonambas:
### What Changed
- Removes spreading of unknown props on Snackbar

### How To Test or Verify
- Verify story renders as expected: http://localhost:9001/?path=/story/feedback-snackbar--responsive-system-props
- Verify tests did not break

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Snackbar/Snackbar.js
+++ b/packages/matchbox/src/components/Snackbar/Snackbar.js
@@ -23,7 +23,7 @@ const StyledClose = styled(Box)`
 `;
 
 const Snackbar = React.forwardRef(function Snackbar(props, userRef) {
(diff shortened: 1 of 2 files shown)


packages/matchbox/src/components/Snackbar/Snackbar.js
```diff
@@ -31,8 +33,8 @@ const Snackbar = React.forwardRef(function Snackbar(props, ref) {
       p="300"
       borderRadius="100"
       ref={ref}
-      {...rest}
       tabIndex="-1"
+      {...systemProps}
```
jonambas, line 37: I think we should support data-id even if the UI doesn't currently use it

logansparlin (approved):
Looks good

PR #766, opened 2021-02-19
UX-395 Remove spreading of unknown Banner props

jonambas:
### What Changed
- Removes spreading of unknown props on Banner

### How To Test or Verify
- Verify story renders as expected: http://localhost:9001/?path=/story/feedback-banner--default-banner
- Verify tests did not break

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Banner/Banner.js
+++ b/packages/matchbox/src/components/Banner/Banner.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { deprecate } from '../../helpers/propTypes';
-import { omit } from '../../helpers/props';
+import { pick, omit } from '../../helpers/props';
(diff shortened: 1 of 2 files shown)


packages/matchbox/src/components/Banner/Banner.js
```diff
@@ -141,7 +142,7 @@ const Banner = React.forwardRef(function Banner(props, ref) {
       py={size === 'large' ? null : '200'}
       borderRadius="100"
       status={status}
-      {...rest}
+      {...systemProps}
       ref={ref}
       tabIndex="-1"
       overflow="hidden"
```
jonambas, line 148: I think we should support data-id even if the UI doesn't currently use it

logansparlin (approved):
Looks good! Checked locally and looks like only props not supported from FE were used to fix muted banners when they were broken

PR #767, opened 2021-02-22
UX-452 Remove spreading of unknown Tag props

jonambas:
### What Changed
- Removes spreading of unknown props on Tag
- Adds data-id and forwardRef

### How To Test or Verify
- Verify story renders as expected: http://localhost:9001/?path=/story/feedback-tag--colors
- Verify tests did not break

### PR Checklist


logansparlin (approved):
Looks good! Everything working locally and no unsupported props from FE

PR #770, opened 2021-02-25
UX-465 Fix empty Panel.Section rendering

jonambas:
### What Changed
- Fixes a bug that prevented empty `Panel.Section` from rendering

### How To Test or Verify
- Run storybook and visit a Panel story: http://localhost:9001/?path=/story/layout-panel--with-a-header
- Try rendering an empty panel section with no children

### PR Checklist


logansparlin (approved):
👍 confirmed it works correctly locally

Issue #205, opened 2019-08-14
The Tag component remove button needs a relevant label for screen reader only users


nicklemmon:
The tag component remove button currently consists of the following markup:

```html
<a class="Tag-module__Close___2BVV6"><svg fill="currentColor" preserveAspectRatio="xMidYMid meet" viewBox="0 0 24 24" width="16" height="16" style="vertical-align: middle;"><g><path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z"></path></g></svg></a>
```

When encountered by screen readers, this will not be interpreted in any meaningful way as all links/buttons need a valid label.

## Potential Solutions

* Leverage a `<ScreenreaderOnly/>` component instance inside the clickable element with the content 'Close' or 'Remove' - see related issue #191 
* Add an `aria-label` to the clickable element, i.e., `<a aria-label="Close"></a>`. The only downside to this approach is that I don't believe `aria-label` content is translated in Chrome...that may have changed since I last heard about that problem though.

PR #771, opened 2021-02-26
UX-428 Update cypress and testing library to latest

jonambas:
### What Changed
Cypress Changelog: https://docs.cypress.io/guides/references/changelog.html#6-0-0

Two main breaking changes that apply to us:
>Cypress now always throws an error when asserting on an element that doesn’t exist in the DOM (unless you’re asserting that the element should not.exist). Assertions such as not.visible, not.contains, not.have.class, etc will now fail when used on non-existent DOM elements. Addresses #205.

> DOM elements where the CSS style (or ancestors) are opacity: 0 are no longer considered visible. However these are still considered actionable and any action commands used to interact with the element will perform the action. This matches browser’s implementation on how they regard elements with opacity: 0. Addresses #4474.

### How To Test or Verify
- Verify integration tests pass
- Verify we are making correct use of `be.visible` and `exist`

### PR Checklist


jonambas:
Need to rebase

logansparlin (approved):
👍 All test passing locally with updated cypress and libby

PR #772, opened 2021-03-01
UX-469 Fix Tabs on Safari

jonambas:
This PR isn't pointed to `main` because we want to publish a patch on top of `4.8.1`. The main branch contains changes that would need a minor release.

### What Changed
- Fixes tab styling on Safari

### How To Test or Verify

- Open Safari and run storybook
- Visit: http://localhost:9001/?path=/story/navigation-tabs--example-tabs
- Verify the first 3 tabs have `type=button`, and the last one does not
- Verify styles are correct

### PR Checklist


logansparlin (approved):
Looks good, fixes styles in safari

PR #773, opened 2021-03-02
UX-470 Fix bug that caused excessive horizontal scroll on Tabs

jonambas:
### What Changed
- Fixes a bug that caused excessive horizontal scroll on the viewport when tabs are overflowing

### How To Test or Verify
- Run storybook and visit: http://localhost:9001/iframe.html?id=navigation-tabs--example-tabs
- Down size the browser, attempt to scroll horizontally with a trackpad on the window
- Verify the browser does **not** scroll horizontally

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Tabs/styles.js
+++ b/packages/matchbox/src/components/Tabs/styles.js
@@ -58,6 +58,7 @@ export const overflowTabs = ({ isOverflowing }) => `
   display: flex;
   visibility: ${isOverflowing ? 'hidden' : 'visible'};
   pointer-events: ${isOverflowing ? 'none' : 'auto'};
+  overflow: ${isOverflowing ? 'hidden' : 'unset'};
 `;
 
 export const containerStyles = () => `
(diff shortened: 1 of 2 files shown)


jonambas:
Feedback from Logan:

- Default to light gray background on right panel (personal preference but I think it looks cleaner)
- Toggles (</> img full screen and external) need to have status indicators
- When clicking on a component, default to the first story instead of requiring two clicks
- Allow code to be shown in a split panel with the preview rather than toggling between code and preview (maybe even allow moving from bottom to right side to full screen)
- Category Sections for components like storybook (media, utility, etc) Icons could be nice
- Close a component dropdown when clicking on another

logansparlin (commented):
I'm not sure if I'm testing wrong but I'm still getting the horizontal scroll

logansparlin (approved):
Nevermind, I'm pretty sure I was testing wrong. Confirmed no horizontal scrolling on the linked story example

PR #786, opened 2021-03-15
UX-457 Remove spreading of unknown Pager props

jonambas:
### What Changed
- Removes spreading of unknown Pager props
- Pager.Next and Pager.Previous are wrappers around Button, so no need to touch those
- Passes through id and data-id

### How To Test or Verify
- Verify no additional props are used in app: https://proply-2web2ui.vercel.app/
- Verify margin props, id, data-id are passed through correctly
- 
### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Pager/Pager.js
+++ b/packages/matchbox/src/components/Pager/Pager.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { margin } from 'styled-system';
 import { createPropTypes } from '@styled-system/prop-types';
+import { pick } from '../../helpers/props';
 import styled from 'styled-components';
 
 import Next from './Next';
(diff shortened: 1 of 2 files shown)


packages/matchbox/src/components/Pager/Pager.js
```diff
@@ -13,9 +14,14 @@ const StyledPager = styled('div')`
 `;
 
 function Pager(props) {
-  const { children, ...rest } = props;
+  const { children, id, 'data-id': dataId, ...rest } = props;
```
logansparlin, line 17: Do we want to allow className here or ok without it?

logansparlin (approved):
Looks good just one small comment

packages/matchbox/src/components/Pager/Pager.js
jonambas: I looked at the UI and looks like we dont need to support it. We should avoid supporting restyling components anyway

PR #794, opened 2021-03-19
UX-474 Allow any Modal children

jonambas:
### What Changed
- Replaces the unnecessary `getChild` or cloning with React context to allow for more flexible component composition

### How To Test or Verify
- Visit http://localhost:9001/?path=Modal__toggle-example&source=false
- Verify clicking on the close button in the header works

### PR Checklist


logansparlin (approved):
Looks good, context looks much better than cloning children

PR #795, opened 2021-03-19
UX-479 Remove spreading of unknown Table component props, Forwards refs to all Table components

jonambas:
### What Changed
- Removes spreading of unknown Table component props
- Supports new HTML attributes for accessibility

### How To Test or Verify
- Run libby, verify the table works as expected http://localhost:9001/?path=Table__all-table-components&source=false
- Verify 2web2ui does not use any attributes that aren't defined 

### PR Checklist

### DIFF
--- a/libby/layout/Table.lib.js
+++ b/libby/layout/Table.lib.js
@@ -24,7 +24,7 @@ describe('Table', () => {
       <Table title="My Table">
         <thead>
           <Table.Row header>
-            <Table.HeaderCell width="20%">Heading 1</Table.HeaderCell>
+            <Table.HeaderCell>Heading 1</Table.HeaderCell>
             <Table.HeaderCell>Heading 2</Table.HeaderCell>
             <Table.HeaderCell>Heading 3</Table.HeaderCell>
           </Table.Row>
@@ -84,7 +84,7 @@ describe('Table', () => {
       <Table>
         <thead>
           <Table.Row header>
-            <Table.HeaderCell>Padding 300</Table.HeaderCell>
+            <Table.HeaderCell width="20%">Padding 300</Table.HeaderCell>
             <Table.HeaderCell>Padding 300</Table.HeaderCell>
             <Table.HeaderCell>Padding 300</Table.HeaderCell>
           </Table.Row>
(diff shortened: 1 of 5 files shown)


packages/matchbox/src/components/Table/TableElements.js
```diff
@@ -20,76 +20,105 @@ const StyledHeaderCell = styled('th')`
 
 const StyledRow = styled('tr')`
   ${row}
-  ${padding}
```
jonambas, line 23: I don't remember why we added this in – padding doesn't work on `tr`

libby/layout/Table.lib.js
```diff
@@ -90,15 +95,15 @@ describe('Table', () => {
           </Table.Row>
         </thead>
         <tbody>
-          <Table.Row p="800">
```
jonambas, line 93: Removed this because padding doesn't work on `tr`. I don't think it ever did, doesn't look like we were overwriting padding context in the row

jonambas:
Need to add a few more properties: https://dequeuniversity.com/library/aria/table-sortable

packages/matchbox/src/components/Table/Table.js
```diff
@@ -51,7 +51,12 @@ function Table(props) {
       onScroll={freezeFirstColumn ? handleScroll : null}
       {...marginProps}
     >
-      <StyledTable freezeFirstColumn={freezeFirstColumn} isScrolled={isScrolled} {...rest}>
+      <StyledTable
```
jonambas, line 54: Add `aria-readyonly` and `role`

packages/matchbox/src/components/Table/TableElements.js
```diff
@@ -20,76 +20,128 @@ const StyledHeaderCell = styled('th')`
...
+      {...paddingContext}
+      {...paddingProps}
+      className={className}
+      fontSize={['200', null, '300']}
+      colSpan={colSpan}
+      ref={userRef}
+      role={role}
+      rowSpan={rowSpan}
+      style={style}
+      width={width}
```
logansparlin, line 49: Should we use system props for this?

logansparlin (approved):
Looks good, just one small question

packages/matchbox/src/components/Table/TableElements.js
jonambas: Yup

PR #801, opened 2021-03-23
Resolves #788 Support labelHidden on Select

### LINKED ISSUE #788: Add `labelHidden` support for the `<Select />` component
## 🚀 Feature request

**Current Behavior**

Now that the old theme in the app has been removed, we can add support for `labelHidden` for the `<Select />` to keep it consistent with `<TextField />`. To add a screen reader only label to a `<Select />`, we end up doing the following:

```jsx
<Select label={<ScreenReaderOnly>Compare By</ScreenReaderOnly>} />
```

This works, but it throws a `PropTypes` error in the console and is an inconsistent API vs. the other form controls.

**Desired Behavior**

Implement the `lableHidden` prop just like the `TextField` component.

**Suggested Solution**

```jsx
<Select label="Compare By" labelHidden />
```

* Improves prop API consistency between components
* Implementation details are handled by the component vs. the engineer working on a feature

jonambas:
Resolves #788 

### What Changed
- Adds `labelHidden` to Select

### How To Test or Verify
- Run libby, visit http://localhost:9001/?path=Select__hidden-label&source=false
- Verify the label is present to screen readers, but visually hidden

### PR Checklist

### DIFF
--- a/cypress/integration/Select.spec.js
+++ b/cypress/integration/Select.spec.js
@@ -13,4 +13,10 @@ describe('The Select component', () => {
     cy.get('[data-id="error-message"]').should('exist');
     cy.findByText('You forgot to select').should('exist');
   });
+
+  it('renders with a hidden label correctly', () => {
+    cy.visit('/iframe.html?path=Select__hidden-label&source=false');
+    cy.get('#idLabel').should('have.text', 'Select an option');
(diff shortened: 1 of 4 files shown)


logansparlin (approved):
looks good

PR #804, opened 2021-03-24
UX-468 Tab focus state uses focus-visible with fallback

jonambas:
### What Changed
- Tab focus state now uses `focus-visible`
- A fallback is provided for browsers that don't support `focus-visible`

### How To Test or Verify
- Run libby with `npm run start:libby`
- Visit http://localhost:9001/?path=Tabs__automatic-keyboard-activation&source=false
- Verify clicking on tabs does not produce tab outlines
- Verify keyboard navigation does produce tab outlines

### PR Checklist


logansparlin (approved):
Looks good

PR #816, opened 2021-04-06
Resolves #810 – Support Label children over label prop

### LINKED ISSUE #810: Change Label component API to render children instead of "label" string prop
## 🚀 Feature request

**Current Behavior**

```jsx
<Label label="First Name" />
```

**Desired Behavior**

```jsx
<Label>First Name</Label>
```

**Suggested Solution**

Introduce a breaking change to the component API to improve the component API, allowing for improved component composition.

This is more of a developer experience improvement than anything else.

**Additional context**

It looks like there aren't any docs for the `<Label />` on design.sparkpost.com. This may be intentional since it isn't used on its own with much frequency.

Thanks 👋

jonambas:
Closes #810

### What Changed
- Removes the check in the Label component that returns `null` if no `label` is provided
- Deprecates `label`
- Updates components that use `Label`
- Installs react-testing-library
- Refactor a few tests

### How To Test or Verify
- Run libby, and visit the following component pages to verify Label works as expected:
- TextField
- Select
- Checkbox
- Radio
- ListBox
- ComboBox

### PR Checklist

### DIFF
--- a/config/test-utils.js
+++ b/config/test-utils.js
@@ -1,0 +1,18 @@
+// // react-testing-library setup
+import React from 'react';
+import { render } from '@testing-library/react';
+import ThemeProvider from '../packages/matchbox/src/components/ThemeProvider/ThemeProvider';
+
+function Wrapper({ children }) {
+  return <ThemeProvider>{children}</ThemeProvider>;
+}
+
+function customRender(ui, options) {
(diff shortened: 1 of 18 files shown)


logansparlin (approved):
looks good!

PR #817, opened 2021-04-08
Resolves #811 Support Label htmlFor

### LINKED ISSUE #811: Update Label component API to accept the `htmlFor` prop rather than deriving its value from `id`
## 🚀 Feature request

**Current Behavior**

```jsx
<Label id="my-label" />
```

Renders as:

```html
<label for="my-label" id="my-labelLabel" />
```

**Desired Behavior**

```jsx
<Label htmlFor="my-form-field" id="my-label" />
```

Renders as:

```html
<label for="my-form-field" id="my-label" />
```

**Suggested Solution**

Remove the derivation of the `htmlFor` prop and just pass the prop.

The current implementation causes a few problems:

- The API is confusing - the rendered DOM is surprising vs. the props passed in
- When using `downshift`, the `getLabelProps` method returned by `useCombobox` passes a value both for `htmlFor` and for `id` - this doesn't work with the current implementation of the component and causes some focus management/a11y bugs
- This _might_ block https://sparkpost.atlassian.net/browse/FE-1386

Thanks again 👋

jonambas:
Closes #811 

### What Changed
- Passes through `htmlFor` and `id` if both is provided
- Defaults to old behavior if only `id` is provided

### How To Test or Verify
- Verify tests pass, we thoroughly test if ids are working properly through tests

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Label/Label.js
+++ b/packages/matchbox/src/components/Label/Label.js
@@ -53,7 +53,7 @@ Label.displayName = 'Label';
 Label.propTypes = {
   label: deprecate(PropTypes.string, 'Use the children instead'),
   children: PropTypes.node,
-  id: PropTypes.string.isRequired,
+  id: PropTypes.string,
   htmlFor: PropTypes.string,
 };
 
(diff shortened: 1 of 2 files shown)


logansparlin (approved):
looks good

PR #825, opened 2021-04-20
UX-507 Pass through all ActionList props for downshift

jonambas:
### What Changed
- Passes through all ActionList props to support downshift

### How To Test or Verify
- Verify non-declared props are passed along, such as `id` `role` or an `aria` attribute

### PR Checklist


logansparlin (approved):
looks good

PR #827, opened 2021-04-22
UX-510 Fix Popover activators being focused by default

jonambas:
### What Changed
- Adjusts the initial incorrect implementation – activators shouldn't be focused when clicking outside

### How To Test or Verify
- Visit http://localhost:9001/?path=Popover__controlled&source=false
- The activator button should not be focused by default
- The activator button should only focus when pressing escape when open

### PR Checklist


logansparlin (approved):
Confirmed locally, focus only happens when escaping out of popover

PR #834, opened 2021-05-04
UX-517 Fix wrapping and alignment on Slider ticks

### LINKED ISSUE #833: `Slider` needs labeling support
## 🚀 Feature request

**Current Behavior**

```jsx
<>
  <label id="my-label">My Slider</label>
  <Slider aria-labelledby="my-label" />
</>
```

`aria-labelledby` does not render on the element with `role="slider"`

**Desired Behavior**

Ideally, the above example would allow association between a slider instance and a label _or_ a `label` and `labelHidden` prop could be added that handle this under the hood:

```jsx
<Slider label="My Slider" labelHidden id="my-slider" />
```

**Suggested Solution**

Pass in `aria-labelledby` but be sure that the attribute renders on the element with `role="slider"`, not the wrapping `<div />`.

There could also potentially be a `label` and `labelHidden` prop added to the component that handle this association under the hood - much like `TextField`.

Working on https://sparkpost.atlassian.net/browse/TR-2852 and the `Slider` instances need labels

jonambas:
Closes #833 

### What Changed
- Slider ticks with spaces should no longer wrap
- Slider ticks and the beginning and end of the slider should no longer overflow
- Slider supports `aria-labelledby`

### How To Test or Verify
- Run libby
- Visit http://localhost:9001/?path=Slider__with-ticks&source=false
- Verify the tick labels dont wrap, and that the first and last tick labels don't overflow (they should be left and right aligned).

### PR Checklist

### DIFF
--- a/libby/form/Slider.lib.js
+++ b/libby/form/Slider.lib.js
@@ -25,8 +25,8 @@ describe('Slider', () => {
         disabled
         value={33}
         ticks={{
-          50: '50',
-          25: 'Recommended',
+          0: '10 days',
+          100: '100000 days',
         }}
       />
     </>
--- a/packages/matchbox/src/components/Slider/Slider.js
+++ b/packages/matchbox/src/components/Slider/Slider.js
@@ -17,6 +17,7 @@ import {
   tick,
   tickHover,
   tickLabel,
+  tickLabelPosition,
   handle,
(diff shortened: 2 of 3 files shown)


logansparlin (approved):
looks good

PR #835, opened 2021-05-04
UX-518 prevent form from submitting when ListBox value is changed

logansparlin:
### What Changed

- Prevent form from submitting when choosing a `ListBox` option
- New test to verify form is not being submitted

### How To Test or Verify

- Verify new example in libby http://localhost:9001/?path=ListBox__in-parent-form&source=false
- Verify cypress tests pass and form is not submitted

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/ListBox/Option.js
+++ b/packages/matchbox/src/components/ListBox/Option.js
@@ -12,6 +12,11 @@ const Option = React.forwardRef(function Option(props, ref) {
     return selected === value;
   }, [selected, value]);
 
+  const handleClick = event => {
+    event.preventDefault();
+    onSelect && onSelect(value);
+  };
+
   return (
     <Box
       as="li"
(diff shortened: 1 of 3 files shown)


packages/matchbox/src/components/ListBox/Option.js
```diff
@@ -21,7 +26,7 @@ const Option = React.forwardRef(function Option(props, ref) {
       aria-posinset={index}
       aria-selected={isActive}
       disabled={disabled}
-      onClick={() => onSelect(value)}
+      onClick={handleClick}
     >
       <StyledLink active={isActive} as="button" disabled={disabled} tabIndex="-1" ref={ref}>
```
jonambas, line 31: I _think_ the issue might be because this button doesn't have `type="button"`, since buttons default to submit

logansparlin: Oh good call

jonambas (approved):
Looks good!

PR #854, opened 2021-05-19
UX-538 Fix controlled Listboxes

jonambas:
### PR Checklist

### DIFF
--- a/libby/form/ListBox.lib.js
+++ b/libby/form/ListBox.lib.js
@@ -163,4 +163,33 @@ describe('ListBox', () => {
       </ListBox>
     </form>
   ));
+
+  add('as a controlled component', () => {
+    const [value, setValue] = React.useState(2);
+
+    function handleClick() {
+      if (value === 4) {
+        setValue(1);
+      } else {
+        setValue(value + 1);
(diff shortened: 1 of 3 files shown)


libby/form/ListBox.lib.js
```diff
@@ -163,4 +163,33 @@ describe('ListBox', () => {
...
+    function handleClick() {
+      if (value === 4) {
+        setValue(1);
+      } else {
+        setValue(value + 1);
+      }
+    }
+    return (
+      <>
+        <ListBox
```
logansparlin, line 179: Should probably `setValue()` on change, values get out of syncwhen using the increment button and then choosing a different option by clicking

logansparlin (changes requested):
Looks good! Just one small change to the example

libby/form/ListBox.lib.js
jonambas: Hmm this is probably a bug with the component – the ticket is to fix the listbox when only value is controlled

jonambas: Oh nvm i see what youre saying

PR #864, opened 2021-06-07
UX-565 Massively change ActionList and Popover interactions

jonambas:
### What Changed
- Changes the behavior of Popovers
  - Popovers are now rendered in a `Portal`
  - Popovers now correctly focus on a focusable child when opening
  - Popovers now open when hitting the up or down arrows when uncontrolled
  - Popovers now close when pressing tab (via a new prop)
- ActionList
  - ActionLists are now keyboard navigable with the up and down arrows
  - ActionLists now properly set `role="menuitem"` and `role="menu"`
- Table
  - Tables now only render one table with the `freezeFirstColumn` prop

### How To Test or Verify

- Run libby
- Visit http://localhost:9001/?path=Popover__with-an-ActionList
  - Verify the changes above
- On: http://localhost:9001/?path=Popover__uncontrolled
  - Verify the popover container is focused when opened
- On: http://localhost:9001/?path=Popover__controlled
  - Verify the close button is focused when opened
- On: http://localhost:9001/?path=Table__frozen-first-column
  - Verify popovers work correctly in the first column of the table

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Table/Table.js
+++ b/packages/matchbox/src/components/Table/Table.js
@@ -9,39 +9,17 @@ import { pick } from '../../helpers/props';
 import { Cell, HeaderCell, Row, TotalsRow } from './TableElements';
 import SortButton from './SortButton';
 import { TablePaddingContext } from './context';
-import { table, wrapper } from './styles';
+import { table, wrapper, sticky } from './styles';
 
 const StyledTable = styled('table')`
   ${table}
   ${padding}
+  ${sticky}
 `;
 
 const Wrapper = styled(Box)`
   ${wrapper}
   ${margin}
-`;
-
-const DuplicatedTable = styled(Box)`
-  pointer-events: none;
-  z-index: ${({ theme }) => theme.zIndices.default};
-
-  th:not(:first-child),
-  td:not(:first-child) {
-    visibility: hidden;
-    pointer-events: none;
-    border-color: transparent !important;
-  }
-  tr {
-    background: none !important;
-  }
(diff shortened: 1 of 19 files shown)


packages/matchbox/src/components/Table/Table.js
```diff
@@ -74,50 +50,33 @@ function Table(props) {
...
+      freezeFirstColumn={freezeFirstColumn}
+      onScroll={freezeFirstColumn ? handleScroll : null}
+      {...marginProps}
+    >
+      <StyledTable
+        aria-readonly={readOnly}
+        data-id={dataId}
+        freezeFirstColumn={freezeFirstColumn}
+        id={id}
+        isScrolled={isScrolled}
```
jonambas, line 67: There is a bug in the current version of chrome that prevents box shadow from rendering I think. Looks fine on other browsers.

jonambas:
- The button needs to have aria-controls with the value as the id on the element with role="menu" (the button controls the menu)
- The button needs to have aria-haspopup="true"
- The element with role="menu" needs to have aria-labelledby with the value as the id of the button element (the menu is labelled by the button)

packages/matchbox/src/components/Table/Table.js
logansparlin: I think you'd have to use display: block or inline-block to get the box-shadow to show up

logansparlin (approved):
looks good!

logansparlin (approved):
lgtm!

PR #867, opened 2021-06-14
UX-568 Use theo instead of gulp-theo

jonambas:
### What Changed
- Removes gulp
- Removes gulp-theo
- Adds theo
- Adds snapshots to verify the change went smoothly

### How To Test or Verify
- This takes a bit of manual work but the best way to verify is to run the snapshot tests between old and new
- Or do a diff of the token files before and after the change
- Also verify the outputted filenames did not change

### PR Checklist


logansparlin (approved):
looks good!

PR #875, opened 2021-06-24
UX-581 Fix Inline child being excluded from Banner component

logansparlin:
### What Changed

- Updated `excludeChild` function to not check `name`, only `displayName`

### How To Test or Verify

- Testing this one is a little challenging, as it was only happening in 2web2ui
- Make sure you're using this version of Matchbox in 2web2ui
- `npm run build` in 2web2ui
- `npm run serve` and check that the `Inline` component isn't being excluded

### PR Checklist


packages/matchbox/src/helpers/children.js
```diff
@@ -25,10 +25,7 @@ export function getChild(name, children, passedProps = {}) {
 
 export function excludeChild(names, children) {
   return React.Children.map(children, child => {
-    if (
-      React.isValidElement(child) &&
-      (names.includes(child.type.displayName) || names.includes(child.type.name))
-    ) {
+    if (React.isValidElement(child) && names.includes(child.type.displayName)) {
```
jonambas, line 28: Should we write a unit test for this?

jonambas (approved):

PR #886, opened 2021-07-02
UX-584 Fix Popover Overlay bug

logansparlin:
### What Changed

- Update positioning values for `PopoverOverlay` when popover is open or closed

### How To Test or Verify

- Verify the Popover [Story](http://localhost:9001/?path=Popover__positioning) 
- Open and close the popover, make the window smaller, and try scrolling to the right. Before the page would overflow because the overlay element positioning had not been updated

### PR Checklist


jonambas (approved):
Nice, glad that was an easy fix

PR #898, opened 2021-08-03
UX-577 Typescript research findings

logansparlin:
### PR Checklist


packages/matchbox/index.d.ts
```diff
@@ -0,0 +1,3 @@
+export * from './components';
```
jonambas, line 1: This doesn't look like a definitions file, should this be `index.ts`?

packages/matchbox/package.json
```diff
@@ -61,7 +64,10 @@
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-terser": "^7.0.2",
-    "styled-components": "^5.2.1"
+    "rollup-plugin-typescript2": "^0.30.0",
+    "styled-components": "^5.2.1",
+    "tslib": "^2.3.0",
+    "typescript": "^4.3.5"
```
jonambas, line 70: Will also need to export definitions from this file `"types": "./types.d.ts"`

jonambas:
wonder if we need to add anything from https://github.com/typescript-eslint/typescript-eslint

packages/matchbox/index.d.ts
logansparlin: Oh right so I just ignored the .d.ts files that were generated just to keep the example PR clean to see the main points. It actually generates a .d.ts file for every component / file and exports to `./components` `./hooks` and `./helpers` so we'd need to wrap that all up in one file or one directory we include so FE has the definitions

packages/matchbox/package.json
logansparlin: Good point

packages/matchbox/index.d.ts
jonambas: Ohhh i see

PR #907, opened 2021-08-11
UX-610 Convert ActionList to typescript

jonambas:
Note: `ts-loader` is extremely slow

### What Changed
- Converts `ActionList` and all its components to typescript
- Fixes `IconBase` types

### How To Test or Verify
- Verify the build works and type definitions for `ActionList` are generated
- Verify libby and the libby build work

### PR Checklist

- [ ] Add the correct `type` label
- [ ] Pull request approval from #uxfe or #design-guild

PR #908, opened 2021-08-12
UX-614 Convert Button to typescript

jonambas:
Requires #907 to be merged first

### What Changed
- Converts `Button` to typescript
- Add a type helper for polymorphic components with `as`

### How To Test or Verify
- Verify libby button stories work
- Verify Button tests pass

### PR Checklist


packages/matchbox/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,335 @@
...
+      onFocus,
+      onBlur,
+      buttonSize,
+      buttonVariant,
+      buttonColor,
+      ref,
+      loading,
+      ...systemProps,
+      ...componentProps,
+    };
```
logansparlin, line 242: Should we add types for these?

packages/matchbox/src/components/Button/Group.tsx
```diff
@@ -4,11 +4,16 @@ import styled from 'styled-components';
...
+  className?: string;
+  children?: React.ReactNode;
+};
+
 const StyledGroup = styled(Box)`
   ${group}
 `;
 
-const Group = ({ children, className = '' }) => (
+const Group: React.FC<GroupProps> = ({ children, className = '' }) => (
```
logansparlin, line 16: I don't think we need React.FC here since you're explicitly typing className and children

logansparlin (approved):
Looks good, just a couple small comments

packages/matchbox/src/components/Button/Button.tsx
jonambas: These are already typed no?

logansparlin: oh yea my bad, not sure what I was thinking 😆

PR #923, opened 2021-08-19
UX-640 Convert Panel to typescript

logansparlin:
### What Changed

- Converts `Panel` to typescript

### How To Test or Verify

- Verify all tests pass and the `Panel` stories are correct

### PR Checklist


packages/matchbox/src/components/Panel/Action.tsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 import { Button } from '../Button';
 
+type ActionProps = React.ComponentProps<typeof Button>;
```
jonambas, line 4: `as` prop isn't working here on `Panel.Action`

jonambas: We should add a libby story to test with a dummy component on `as`

packages/matchbox/src/helpers/types.ts
```diff
@@ -60,4 +60,15 @@ type AlignY = 'center' | 'top' | 'bottom';
 
 type Breakpoints = 'default' | 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 
-export type { ForwardRefComponent, OwnProps, IntrinsicElement, Merge, AlignX, AlignY, Breakpoints };
+type Headings = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
jonambas, line 63: Nice

jonambas (changes requested):
One change needed for the `as` prop on `Panel.Action`

jonambas (approved):

PR #937, opened 2021-08-24
UX-642 convert Popover to typescript

logansparlin:
### What Changed

- Converts `Popover` to typescript

### How To Test or Verify

- Verify all tests pass
- Verify libby stories

### PR Checklist


packages/matchbox/src/components/Popover/Popover.tsx
```diff
@@ -1,16 +1,68 @@
...
+    top?: boolean;
+    /**
+     * @deprecated Use `position` instead
+     */
+    bottom?: boolean;
+    position?: ResponsiveValue<'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight'>;
+    /**
+     * Callback function that is called when clicking outside the popover, or hitting escape.
+     */
+    onClose?: (any) => void;
```
jonambas, line 49: Looks like this could either be a keydown or click event. I think `(e: Event) => void;` would work

packages/matchbox/src/components/Popover/Popover.tsx
```diff
@@ -80,9 +132,9 @@ const Popover = React.forwardRef(function Popover(props, ref) {
   function handleOutsideClick(e) {
```
jonambas, line 132: `Parameter 'e' implicitly has an 'any' type`

Was trying to add types but not sure what it should be, probably fine without it though

packages/matchbox/src/components/Popover/PopoverContent.tsx
```diff
@@ -66,21 +67,31 @@ function getPositionFromDeprecatedProps({ top, left }) {
...
 }
 
-const Content = React.forwardRef(function Content(props, userRef) {
+type ContentProps = PaddingProps &
+  LayoutProps &
+  Pick<
+    React.ComponentProps<typeof Popover>,
+    'children' | 'top' | 'left' | 'position' | 'className' | 'sectioned' | 'open'
+  > & {
+    style?: any;
```
jonambas, line 76: I think `React.CSSProperties` works here

jonambas (changes requested):

PR #940, opened 2021-08-26
UX-659 Convert Tooltip to typescript

jonambas:
### What Changed
- Converts Tooltip to typescript

### How To Test or Verify
- Verify the tooltip stories work correctly

### PR Checklist


packages/matchbox/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,29 +1,75 @@
...
+}>`
   ${visibility}
 `;
 
 const StyledContent = styled('div')`
   ${system}
 `;
 
-function Tooltip(props) {
+type TooltipProps = {
```
logansparlin, line 40: Should export

packages/matchbox/src/components/Tooltip/Tooltip.tsx
```diff
@@ -151,59 +196,4 @@ function Tooltip(props) {
...
-  ...createPropTypes(typography.propNames),
-  ...createPropTypes(shadow.propNames),
-};
-
-Tooltip.defaultProps = {
-  right: true,
-  bottom: true,
-  horizontalOffset: '0px',
-  forcePosition: false,
-  preferredDirection: {
```
logansparlin, line 201: Do we need to update these defaults in the props spread?

jonambas: these actually weren't used at all, i think we just forgot to remove them at some point

PR #954, opened 2021-09-07
UX-634 Convert ListBox to typescript

logansparlin:
### What Changed

- Converts `ListBox` component to typescript

### How To Test or Verify

- Verify tests pass and stories are correct

### PR Checklist

### DIFF
--- a/libby/form/ListBox.lib.tsx
+++ b/libby/form/ListBox.lib.tsx
@@ -114,9 +114,9 @@ describe('ListBox', () => {
 
   add('with a custom ref', () => {
     function Example() {
-      const ref = React.useRef();
+      const ref = React.useRef<HTMLInputElement>();
       React.useEffect(() => {
-        (ref as React.MutableRefObject<HTMLDivElement>).current.focus();
+        ref.current.focus();
       }, []);
       return (
         <ListBox
--- a/libby/form/ListBox.lib.tsx
+++ b/libby/form/ListBox.lib.tsx
@@ -10,7 +10,7 @@ describe('ListBox', () => {
       defaultValue="option-1"
       label="Select an option"
       data-track="true"
-      onChange={e => console.log(e)}
+      onChange={(e) => console.log(e)}
       onFocus={() => console.log('focus')}
       onBlur={() => console.log('blur')}
     >
@@ -116,14 +116,14 @@ describe('ListBox', () => {
     function Example() {
       const ref = React.useRef();
       React.useEffect(() => {
-        ref.current.focus();
+        (ref as React.MutableRefObject<HTMLDivElement>).current.focus();
       }, []);
       return (
         <ListBox
           id="listbox-1"
           defaultValue="option-2"
           label="Select an option"
-          onChange={e => console.log(e.currentTarget.value)}
+          onChange={(e) => console.log(e.currentTarget.value)}
           ref={ref}
         >
           <ListBox.Option value="option-1">Option 1</ListBox.Option>
@@ -153,7 +153,7 @@ describe('ListBox', () => {
         id="listbox-5"
         label="Select an option"
         placeholder="Select One"
-        onChange={e => console.log(e)}
+        onChange={(e) => console.log(e)}
         name="listbox-form-test"
       >
         <ListBox.Option value="option-1">Option 1</ListBox.Option>
@@ -181,7 +181,7 @@ describe('ListBox', () => {
           label="Select an option"
           placeholder="Select One"
           value={`option-${value}`}
-          onChange={e => setValue(Number(e.currentTarget.value.split('-').pop()))}
+          onChange={(e) => setValue(Number(e.currentTarget.value.split('-').pop()))}
           name="listbox-form-test"
         >
           <ListBox.Option value="option-1">Option 1</ListBox.Option>
--- a/packages/matchbox/src/components/ListBox/ListBox.tsx
+++ b/packages/matchbox/src/components/ListBox/ListBox.tsx
@@ -95,7 +95,7 @@ export type ListBoxProps = {
 } & MarginProps &
   MaxWidthProps &
   MaxHeightProps &
-  React.ComponentPropsWithRef<'input'>;
+  React.ComponentPropsWithoutRef<'input'>;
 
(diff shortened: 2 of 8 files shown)


libby/form/ListBox.lib.tsx
```diff
@@ -116,14 +116,14 @@ describe('ListBox', () => {
     function Example() {
       const ref = React.useRef();
```
jonambas, line 117: You can do `const ref = React.useRef<HTMLInputElement>();` here instead of specifying `React.MutableRefObject`

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -28,40 +34,77 @@ const StyledWrapper = styled(Box)`
...
+  props,
+  userRef,
+) {
   return (
     <StyledButton {...props} ref={userRef}>
       {props.children}
     </StyledButton>
   );
-});
+}) as React.ForwardRefExoticComponent<StyledButtonProps>;
```
jonambas, line 69: Don't think you need this line unless you're extending `ForwardRefExoticComponent`

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -28,40 +34,77 @@ const StyledWrapper = styled(Box)`
...
+  labelHidden?: boolean;
+  helpText?: React.ReactNode;
+  error?: string;
+  errorInLabel?: boolean;
+  disabled?: boolean;
+  required?: boolean;
+  name?: string;
+  optional?: boolean;
+  value?: string | number;
+  onChange?: (ChangeObject) => any;
```
jonambas, line 100: should return `void` right?

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -28,40 +34,77 @@ const StyledWrapper = styled(Box)`
...
+    value?: string | number;
+  };
+};
+
+export type ListBoxProps = {
+  'data-id'?: string;
+  'data-sensitive'?: string;
+  'data-track'?: string;
+  placeholder?: string;
+  id?: string;
```
jonambas, line 89: I think some of these props are already provided by `React.ComponentPropsWithRef<'input'>;` so you might not need to define all of them

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -133,17 +176,17 @@ const ListBox = React.forwardRef(function ListBox(props, userRef) {
     setCurrentValue(value);
 
     // Returns focus to the button when closing the popover
-    buttonRef.current.focus();
+    (buttonRef as React.MutableRefObject<HTMLDivElement>).current.focus();
```
jonambas, line 179: You can do `const buttonRef = React.useRef<HTMLButtonElement>();` instead

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -219,22 +261,23 @@ const ListBox = React.forwardRef(function ListBox(props, userRef) {
...
             </ListBoxButton>
-            <StyledChevron size={24} disabled={disabled} />
+            <StyledChevron size={24} $disabled={disabled} />
           </Box>
         }
       >
         <StyledList
           as="ul"
-          aria-activedescendant={currentValue}
+          aria-activedescendant={currentValue?.toString()}
```
jonambas, line 276: I don't think we have the optional-chaining plugin enabled do we? How is this working 🤔

jonambas: unless it gets included with `@babel/env`?

packages/matchbox/src/components/ListBox/Option.tsx
```diff
@@ -16,36 +25,26 @@ const Option = React.forwardRef(function Option(props, ref) {
...
+        $disabled={disabled}
+        tabIndex={-1}
         ref={ref}
       >
         {children}
       </StyledLink>
     </Box>
   );
-});
+}) as React.ForwardRefExoticComponent<OptionProps>;
```
jonambas, line 46: Same here, don't need this line

packages/matchbox/src/components/ListBox/Option.tsx
```diff
@@ -1,11 +1,20 @@
...
 import { Box } from '../Box';
 
-const Option = React.forwardRef(function Option(props, ref) {
+type OptionProps = {
+  selected?: string;
+  value?: string | number;
+  disabled?: boolean;
+  index?: number;
+  setSize?: number;
+  onSelect?: (any) => any;
```
jonambas, line 13: i think `onSelect?: (value: string | number) => void;`  works

packages/matchbox/src/components/ListBox/ListBox.tsx
logansparlin: I'm pretty sure it's enabled since babel 7.8.0

packages/matchbox/src/components/ListBox/ListBox.tsx
```diff
@@ -28,40 +34,77 @@ const StyledWrapper = styled(Box)`
...
+  required?: boolean;
+  name?: string;
+  optional?: boolean;
+  value?: string | number;
+  onChange?: (ChangeObject) => any;
+  defaultValue?: string | number;
+} & MarginProps &
+  MaxWidthProps &
+  MaxHeightProps &
+  React.ComponentPropsWithRef<'input'>;
```
jonambas, line 105: Should be `WithoutRef` i think, since the ref comes from `forwardRef`

packages/matchbox/src/components/ListBox/useOptionConstructor.tsx
```diff
@@ -3,14 +3,30 @@ import { onKey } from '../../helpers/keyEvents';
 
 import Option from './Option';
 
-function useOptionConstructor({ options, value, onSelect, open, placeholder }) {
+type UseOptionConstructorReturnType = {
+  optionsMarkup?: React.ReactNode;
+  focusContainerProps?: {
+    onKeyDown?: (KeyboardEvent) => void;
```
jonambas, line 9: Think you meant `(e: KeyboardEvent)`

jonambas (approved):
looks good!

PR #956, opened 2021-09-08
UX-648 Convert Select to typescript

logansparlin:
### PR Checklist


packages/matchbox/src/components/Select/Select.tsx
```diff
@@ -87,15 +63,28 @@ const StyledSelect = styled(SelectBox)`
...
+  helpText?: React.ReactNode;
+  error?: string;
+  errorInLabel?: boolean;
+  optional?: boolean;
+  placeholderValue?: string;
+} & MarginProps &
+  MaxWidthProps &
+  React.ComponentPropsWithoutRef<'input'>;
+
+const Select = React.forwardRef<HTMLInputElement, SelectProps>(function Select(props, userRef) {
```
jonambas, line 87: Should this be `HTMLSelectElement`?

packages/matchbox/src/components/Select/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as Select } from './Select';
+export * from './Select';
+export * from './Options';
+export * from './Option';
```
jonambas, line 4: Do we need to export the types? We don't expose the `Option` and `Options` components to end users

jonambas (approved):
LGTM

PR #960, opened 2021-09-09
UX-645 convert Radio to typescript

logansparlin:
### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/Radio/Radio.tsx
+++ b/packages/matchbox/src/components/Radio/Radio.tsx
@@ -24,7 +24,8 @@ export type RadioProps = {
   onFocus?: React.FocusEventHandler;
   onBlur?: React.FocusEventHandler;
   helpText?: React.ReactNode;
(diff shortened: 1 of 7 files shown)


packages/matchbox/src/components/Radio/Radio.tsx
```diff
@@ -53,7 +66,6 @@ const Radio = React.forwardRef(function Radio(props, userRef) {
           onFocus={onFocus}
           onBlur={onBlur}
           type="radio"
-          error={error}
           ref={userRef}
           {...describedBy}
           {...componentProps}
```
jonambas, line 71: Because we spread props, I think we'll need to include `React.ComponentPropsWithoutRef<'input'>`. And maybe omit `type` from it?

jonambas (changes requested):
Need to pull in other `<input>` props I think onto the Radio component

jonambas (approved):
LGTM

PR #962, opened 2021-09-10
UX-660 Convert TimelineTabs to typescript

jonambas:
### What Changed
- Convert TimelineTabs into typescript

### How To Test or Verify
- Verify the TimelineTabs stories work as expected

### PR Checklist


packages/matchbox/src/components/TimelineTabs/useTabConstructor.tsx
```diff
@@ -1,9 +1,16 @@
 import React from 'react';
 import { onKey, onKeys } from '../../helpers/keyEvents';
 
-function useTabConstructor({ tabs, initialIndex }) {
-  const tabRefs = React.useRef({ current: [] });
-  const focusRef = React.useRef();
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function useTabConstructor({ tabs, initialIndex }: { tabs: any[]; initialIndex: number }): {
```
jonambas, line 5: Not sure what this should be, might need to refactor a bit to fix this `any[]`

packages/matchbox/src/components/TimelineTabs/TimelineTabs.tsx
```diff
@@ -11,7 +9,16 @@ const Container = styled.div`
   ${margin}
 `;
 
-const TimelineTabs = React.forwardRef(function TimelineTabs(props, userRef) {
+export type TimelineTabsProps = {
+  children?: React.ReactNode;
+  id?: string;
```
logansparlin, line 14: Don't necessarily need this `id` type since using `HTMLDivElement`

packages/matchbox/src/components/TimelineTabs/useTabConstructor.tsx
logansparlin: Would `React.ComponentProps<typeof TimelineTabsTab>[]` work?

logansparlin (approved):
looks good, just a couple small comments. Nothing blocking

packages/matchbox/src/components/TimelineTabs/useTabConstructor.tsx
jonambas: It's technically a `ReactChild` that is returned from `React.Children.toArray` I think, but it breaks because of the `...tab.props,` later in the file

PR #978, opened 2021-09-14
UX-679 convert event helpers to typescript

logansparlin:
### PR Checklist


packages/matchbox/src/helpers/event.ts
```diff
@@ -0,0 +1,35 @@
...
+      func.apply(context, args);
+    }
+  };
+}
+
+export function noop(): void {
+  return undefined;
+}
+
+export function identity<T>(a: T) {
```
jonambas, line 29: Nice 👍

packages/matchbox/src/helpers/event.ts
```diff
@@ -0,0 +1,35 @@
...
+  return a;
+}
+
+export function isNotTouchEvent(e: TouchEvent): boolean {
```
jonambas, line 33: Not sure if the React version of this type is compatible/required here, but the type used by `onTouchStart` is `onTouchStart?: React.TouchEventHandler<HTMLDivElement>`

jonambas: Maybe `isNotTouchEvent<T extends HTMLElement>(e: React.TouchEventHandler<T>)`

Up to you though, looks like the build passes anyway

jonambas (approved):

packages/matchbox/src/helpers/event.ts
jonambas: Oh wait i think i gave you bad advice, this is the even object not the event handler

PR #983, opened 2021-09-15
UX-684 Convert prop helpers to typescript

jonambas:
### PR Checklist

### DIFF
--- a/packages/matchbox/src/helpers/props.ts
+++ b/packages/matchbox/src/helpers/props.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 /**
  * Omits styled system props from an object
  * @param {Object} props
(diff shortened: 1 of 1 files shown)


packages/matchbox/src/helpers/props.ts
```diff
@@ -9,11 +10,11 @@
  * omit({ my: '100', className: 'classs' }, margin.propNames)
  *  > { className: 'class' }
  */
-export function omit(props, names) {
+export function omit(props: { [k: string]: any }, names: string[]): { [k: string]: any } {
```
logansparlin, line 13: shoudl we use unknown instead of any?

logansparlin (approved):
looks good, just one small comment

packages/matchbox/src/helpers/props.ts
jonambas: Yeah, I tried any it was breaking a few files

PR #991, opened 2021-09-20
UX-702 ThemeProvider target should not be required

jonambas:
To reproduce the bug:
- Open up a Typescript project (Or a new one with `npx create-next-app test-project --ts`
- Try using `ThemeProvider` without the `target` prop

### What Changed
- Makes ThemeProvider target optional

### How To Test or Verify
- No easy way to verify this one since it cannot be reproduced in Libby
- Possibly running the build, copying over types from `dist` into `node_modules` in a new project

### PR Checklist


jonambas:
Not sure if this is needed anymore because I can't reproduce the error

logansparlin (approved):
looks good

PR #1000, opened 2021-09-23
issue-920 Add closeOnInsideClick prop to Popover

jonambas:
### PR Checklist

### DIFF
--- a/packages/matchbox/src/helpers/keyEvents.ts
+++ b/packages/matchbox/src/helpers/keyEvents.ts
@@ -1,3 +1,5 @@
+import React from 'react';
+
 type EventKeys =
   | 'backspace'
   | 'enter'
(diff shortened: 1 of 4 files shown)


packages/matchbox/src/helpers/keyEvents.ts
```diff
@@ -107,11 +110,11 @@ export function onKey<T extends Element>(event: EventKeys, callback: React.Keybo
  * Multiple key down event handler
  * @example onKeys(['escape', 'enter'], () => foo())(e)
  */
-export function onKeys<T extends Element>(
+export function onKeys(
   events: EventKeys[],
-  callback: React.KeyboardEventHandler,
+  callback: (e: React.KeyboardEvent | KeyboardEvent) => void,
```
logansparlin, line 115: What's the reason for doing this instead of using KeyboardEventHandler?

logansparlin (commented):
Looks good but should add an example in libby

packages/matchbox/src/helpers/keyEvents.ts
logansparlin: Nevermind, I see why you did this in UX-705

PR #1024, opened 2021-11-16
UX-745 Add CRA template

jonambas:
### What Changed
- Adds a CRA template
- See https://create-react-app.dev/docs/custom-templates/

### How To Test or Verify

- Pull down this PR
- Run `npx create-react-app my-app --template file:.path/to/matchbox/packages/matchbox-cra`
- Or run `npx create-react-app my-app --template @sparkpost/matchbox`

### PR Checklist


packages/cra-template-matchbox/template/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
logansparlin, line 3: Could Change this name to something like SparkPost CRA Template

logansparlin (approved):
This looks good. I'm wondering if we should add a .eslintrc but other than that it's good to go

PR #1031, opened 2021-12-02
UX-732 Create new matchbox-css library

logansparlin:
### What Changed

- Adds new @sparkpost/matchbox-css library
- Adds css classes for font-size which handles media queries and various interactive states

### How To Test or Verify

- Run `npm run compile` locally and verify that all the necessary classes were created correctly using the font-size tokens

### PR Checklist

### DIFF
--- a/packages/matchbox-css/package.json
+++ b/packages/matchbox-css/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@sparkpost/matchbox-css",
+  "version": "1.0.0",
+  "description": "Matchbox CSS Utility Library",
+  "main": "index.js",
+  "style": "dist/index.css",
+  "author": "SparkPost",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/SparkPost/matchbox.git"
+  },
+  "scripts": {
+    "prebuild": "npm run clean",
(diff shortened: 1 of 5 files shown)


packages/matchbox-css/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "build": "npm run compile -- --output-style compressed",
+    "clean": "rimraf ./dist"
+  },
+  "dependencies": {
+    "@sparkpost/design-tokens": "^6.1.0"
+  },
+  "devDependencies": {
+    "node-sass": "^6.0.1",
+    "node-sass-import": "^2.0.1",
+    "prettier": "^2.5.0",
```
jonambas, line 28: Can remove prettier if we're not using it

packages/matchbox-css/tokens/font-size.scss
```diff
@@ -0,0 +1,26 @@
+@import '@sparkpost/design-tokens/dist/media-query.map.scss';
```
jonambas, line 1: This works for our existing maps, but the new style dictionary maps are all combined into one. Not sure how much work is involved there, but I think we can merge this in for now and address it later

jonambas (approved):
Couple comments, but we can merge this just to keep the ball rolling

packages/matchbox-css/tokens/font-size.scss
logansparlin: Yea I'll have to update to read from the new `tokens.scss` file from the updated tokens

PR #1035, opened 2021-12-08
UX-734, UX-736, UX-737, UX-738, UX-739, UX-740, UX-741 Token classes

logansparlin:
### What Changed

- Adds typography classes
- Adds elevation classes
- Adds motion classes
- Adds z-index classes
- Adds border classes

### How To Test or Verify

- `npm run compile` locally and verify the classes are created correctly

### PR Checklist

### DIFF
--- a/packages/matchbox-css/index.scss
+++ b/packages/matchbox-css/index.scss
@@ -4,6 +4,15 @@ $interactions: "hover", "focus", "focus-visible", "active", "visited", "disabled
 $media-query-map: map-get($matchbox-tokens, "media-query");
 
 @import './tokens/font-size.scss';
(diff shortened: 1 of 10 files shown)


jonambas (approved):
This looks really good!

Some housekeeping stuff we'll need to take care of, maybe not in this ticket
- Need to remove `"main": "index.js",` in package.json
- Need to specify `"style": "dist/index.css` (i think that key is right)
- Need to specify `"files"`
- Need to add an `.npmignore`

Another thing, but not sure how to really do this, but maybe we should think about how we can test the output. Maybe we create a libby story that uses some classes and snapshot it with Percy?

PR #1041, opened 2021-12-16
UX-759 Add transpileOnly ts-loader flag for development

jonambas:
### What Changed
- Adds the `transpileOnly` flag to libby's ts-loader config to massively speed up TS compilation
- This means that type-checking will not happen when running Libby locally, but we will still have type-checking through the Github workflow (which runs a production build)

### How To Test or Verify
- Run `npm run start:libby` 
- Start up time should be noticeably faster

### PR Checklist

### DIFF
--- a/libby.config.js
+++ b/libby.config.js
@@ -31,6 +31,10 @@ module.exports = {
             exclude: /node_modules/,
             use: {
               loader: 'ts-loader',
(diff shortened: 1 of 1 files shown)


logansparlin (approved):
lgtm

PR #1042, opened 2021-12-16
UX-717 Adds new z-index token, Fix textfield z-indexing with error outlines

jonambas:
### What Changed
- Fixes the right border of the textfield with right-connected components

### How To Test or Verify
- Visit: http://localhost:9001/?path=Visual-Regression__TextField
- Test focus styles on the last textfield with an error

Screen Shot 2021-12-16 at 8 53 30 AM

Screen Shot 2021-12-16 at 8 53 38 AM

### PR Checklist


packages/matchbox/src/components/Connect/Connect.tsx
```diff
@@ -3,10 +3,13 @@ import { Box } from '../Box';
...
 
-const FocusHandler = styled(Box)`
+// $zIncrement is here to ensure the middle element appears
+// above side elements until it side elements are focused
+const FocusHandler = styled(Box)<{ $zIncrement: number }>`
+  z-index: ${(props) => tokens.zIndex_default + props.$zIncrement};
   &:focus-within {
     position: relative;
-    z-index: ${tokens.zIndex_default};
+    z-index: ${tokens.zIndex_default + 2};
```
logansparlin, line 12: This works but seems a little odd, maybe we need to consider a new zIndex token?

logansparlin (approved):
Looks good! Just one small comment

packages/matchbox/src/components/Connect/Connect.tsx
jonambas: would `above` work? with a value of 10 just to be safe?

logansparlin: Yea let's do that

PR #1043, opened 2021-12-16
UX-760 Fix button `to`  types

jonambas:
### What Changed
- Fixes the `to` prop's type so that it accepts objects for react-router support

### How To Test or Verify
- Try to use an object within Button's `to` prop, verify it does not produce an error

### PR Checklist


logansparlin (approved):
nice, looks good

PR #1065, opened 2022-02-18
UX-777 Add small radio card variant

jonambas:
### What Changed
- Adds new `size` prop to `RadioCard` which accepts `small`
- Adds new `space` prop to RadioCard.Group` which accepts `compact`

### How To Test or Verify
- Run Libby, visit http://localhost:9001/?path=RadioCard__small-variant
- Verify the component matches the mockups: https://www.figma.com/file/kbeuqghvpagMmaLWwqszQo/Forms?node-id=260%3A268

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/RadioCard/RadioCard.tsx
+++ b/packages/matchbox/src/components/RadioCard/RadioCard.tsx
@@ -27,7 +27,7 @@ const RadioCard = React.forwardRef<HTMLInputElement, RadioCardProps>(function Ra
     onChange,
     onFocus,
     onBlur,
-    size,
+    size = 'small',
     value,
     weight = 'light',
     'data-id': dataId,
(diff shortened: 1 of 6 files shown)


packages/matchbox/src/components/RadioCard/RadioCard.tsx
```diff
@@ -51,28 +54,35 @@ const RadioCard = React.forwardRef<HTMLInputElement, RadioCardProps>(function Ra
...
               <StyledChecked size="1rem" as={RadioButtonChecked} />
             </Box>
           </Box>
           <Box flex="1" pl="300">
             <StyledHeader
               data-id="radio-card-header"
-              fontSize={fontSize}
-              lineHeight="400"
-              fontWeight="semibold"
+              fontSize={isSmall ? '200' : fontSize}
```
logansparlin, line 70: Maybe we rethink how fontSize is being defined, having `'200' | fontSize` is confusing

logansparlin (approved):
Looks good! Just one small comment

packages/matchbox/src/components/RadioCard/RadioCard.tsx
jonambas: @logansparlin Do you think I should just remove the `weight` prop? We don't use it anywhere in the app, its not in the figma component, and not entirely sure why we set it up like this

logansparlin (approved):
Looks good!

PR #1067, opened 2022-02-28
UX-253 Add CheckboxCard component

jonambas:
### What Changed
- Adds `CheckboxCard` and `CheckboxCard.Group` components

### How To Test or Verify
- Run libby and visit: http://localhost:9001/?path=CheckboxCard__basic-usage
- Verify all the stories work correctly
- Verify the component matches the mockups: https://www.figma.com/file/kbeuqghvpagMmaLWwqszQo/Forms?node-id=279%3A269

### PR Checklist

### DIFF
--- a/packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
+++ b/packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
@@ -61,7 +61,7 @@ const CheckboxCard = React.forwardRef<HTMLInputElement, CheckboxCardProps>(funct
               <StyledChecked size="1.25rem" as={CheckBox} />
             </Box>
           </Box>
-          <Box flex="1" pl="300">
+          <Box flex="1" pl={isSmall ? '400' : '300'}>
             <StyledHeader
               data-id="radio-card-header"
               fontSize={isSmall ? '200' : '300'}
@@ -76,9 +76,10 @@ const CheckboxCard = React.forwardRef<HTMLInputElement, CheckboxCardProps>(funct
           <Box
             data-id="checkbox-card-content"
             pt={isSmall ? '0' : '200'}
-            pl={isSmall ? '450' : '0'}
-            fontSize={isSmall ? '100' : '300'}
-            lineHeight={isSmall ? '100' : '300'}
+            pl={isSmall ? '500' : '0'}
+            fontSize={isSmall ? '200' : '300'}
+            lineHeight={isSmall ? '200' : '300'}
+            color={isSmall ? 'gray.700' : 'gray.900'}
           >
             {children}
           </Box>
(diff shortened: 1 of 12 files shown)


packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+        ref={userRef}
+        type="checkbox"
+        value={value}
+        {...rest}
+      />
+      <StyledLabel as="label" htmlFor={id} $size={size}>
+        <Box display="flex">
+          <Box flex="0 0 auto" width={isSmall ? '200' : '400'}>
+            <Box position="absolute" top={isSmall ? '100' : '500'} left={isSmall ? '200' : '500'}>
+              <StyledUnchecked size="1rem" as={CheckBoxOutlineBlank} />
```
logansparlin, line 60: The sizing and spacing here is wrong. The checkbox on the large card is currently 12px x 12px and should be 16x16, and the spacing to the right of the checkbox should be 12px but it looks a little off

packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+              <StyledChecked size="1rem" as={CheckBox} />
+            </Box>
+          </Box>
+          <Box flex="1" pl="300">
+            <StyledHeader
+              data-id="radio-card-header"
+              fontSize={isSmall ? '200' : '300'}
+              lineHeight={isSmall ? '200' : '400'}
```
logansparlin, line 68: lineHeight here should be `300` not `400`

packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+            </StyledHeader>
+          </Box>
+        </Box>
+        {children && (
+          <Box
+            data-id="checkbox-card-content"
+            pt={isSmall ? '0' : '200'}
+            pl={isSmall ? '450' : '0'}
+            fontSize={isSmall ? '100' : '300'}
+            lineHeight={isSmall ? '100' : '300'}
```
logansparlin, line 81: I think lineHeight and fontSize here when small should be `200`

logansparlin (changes requested):
A few styling comments but overall looks good. Might be for another ticket but there's probably some cleanup and refactoring that could be done to minimize the duplication between parts of the small CheckboxCard and the small RadioCard

packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
jonambas: Hmm looks like the react icons have 4px padding built in

packages/matchbox/src/components/CheckboxCard/CheckboxCard.tsx
jonambas: Looks like this is sent to 400 to account for the extra padding in the icons, same as RadioCard

jonambas: Its a bit offset without it
Screen Shot 2022-03-02 at 9 01 08 AM

logansparlin (approved):
Looks good, matches the mocks now and the spacing feels a lot better

PR #1088, opened 2022-06-16
UX-789 Replace Lerna with Changesets & NPM Workspaces

jonambas:
**Changes**
- Removes lerna and all related dependencies, scripts
- Adds NPM workspaces and updates all scripts
- Reconfigures stale eslint rules to a more recent baseline
- Installs changesets, adds documentation
- Adds turborepo
- (To be published) Adds support for React 18, and updates a few types

**Notes**
- I couldn't upgrade to React 18 because of the dependency on enyzme, which doesn't even officially support React 16. We need to replace it with RTL
- Releases will be a bit different, instead of a single release tag, each package will have its own git tag and release notes separately
- Removes stylelint, I dont think it was even working properly

--- 

**TODOs**
- [x] Find a solution for enzyme with React 18
  - [ ] Use old root rendering from React 17 in Libby 
  - [ ] Alias React 18 for Libby only
  - [ ] Convert enzyme tests to RTL
  - [x] Support both root rendering APIs in Libby and downgrade to 17 here
- [x] Reconfigure eslint rules
- [x] Fix Cypress tab plugin
- [x] Install changesets and document new publishing process


package.json
```diff
@@ -84,63 +67,54 @@
...
+  "overrides": {
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
+  },
+  "workspaces": [
+    "packages/design-tokens",
+    "packages/matchbox-css",
+    "packages/matchbox-icons",
+    "packages/matchbox-media",
+    "packages/matchbox"
```
jonambas, line 114: Intentionally left out the cra and next templates since they are a bit unique and don't have a standard package structure

PR #1097, opened 2022-08-19
Updated RadioCard component and content to have 100% height

llacava:
### What Changed

- The height of radio card component and to 100% and also from its content 
- In a specific use case from subaccount stuff we got this result currently: 

image

vs the expected one: 
image


nstanard (approved):
LGTM!

PR #1098, opened 2022-08-24
build: fix type compilation

jonambas:
### What Changed
- Temporarily removes the `terser` plugin. Was causing a `Error: kill EPERM` error. Need to find an alternative.
- Fixes the types compiled from the build scripts by changing tsconfig's base url. They were missing before. No idea why this works.

### How To Verify
- Run `npm run build`
- Verify that the types are generated in `the dist` folders. Only affects `matchbox` and `matchbox-icons`


logansparlin (approved):
looks good and seems to fix the build

PR #1124, opened 2025-07-08
TASK-10510: Bumping up version of NodeJS to 22.x

dkoerichbird:
This package is a dependency for [support-docs](https://github.com/SparkPost/support-docs) to build, and like there, I am bumping the version of NodeJS here. However, since I am not familiar with this development environment, I focused on building the libraries only. Since there is no history of previous runs of GH actions, I don't know if the checks that are failing here were already failing before.


viniciusgiles:
>Since there is no history of previous runs of GH actions, I don't know if the checks that are failing here were already failing before.

@dkoerichbird The checks passed in your first commits in this PR, but now are failing after the version bumps and other changes. Let's make sure that all checks pass before merging.

PR #1125, opened 2025-07-09
TASK-10510: Bump version of Node.js to 22.x

dkoerichbird:
This package is a dependency for [support-docs](https://github.com/SparkPost/support-docs), and like there, I am bumping the version of Node.js here.


dkoerichbird:
I am withdrawing this PR since the issue with the `support-docs` could be solved. But I feel we have a schedule encounter with this subject in the future...

